PR #679, opened 2014-09-30
Allow record sliding in page.resources.fal

pkerling:
Page\Resources\FalViewHelper was until now missing the feature to use record sliding for page-related media. That is however commonly required for dynamic page background images and such and I wanted to avoid falling back to TypoScript.

It made quite a challenge to separate the already existing sliding code from AbstractContentViewHelper into a separate class and I'm not sure whether the result is quite right. If you have suggestions concerning the implementation of this feature, I'd be happy to rework it.

### DIFF
--- a/Classes/ViewHelpers/AbstractSlideViewHelper.php
+++ b/Classes/ViewHelpers/AbstractSlideViewHelper.php
@@ -1,0 +1,137 @@
+<?php
+namespace FluidTYPO3\Vhs\ViewHelpers;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2014 Philipp Kerling <pkerling@casix.org>
+ *
+ *  All rights reserved
+ *
+ *  This script is part of the TYPO3 project. The TYPO3 project is
+ *  free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  The GNU General Public License can be found at
+ *  http://www.gnu.org/copyleft/gpl.html.
+ *
+ *  This script is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  This copyright notice MUST APPEAR in all copies of the script!
+ * ************************************************************* */
+
+/**
+ * @author Philipp Kerling <pkerling@casix.org>
+ * @package Vhs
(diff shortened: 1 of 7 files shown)


NamelessCoder:
Once again, sorry about the delay @pkerling - I'm a little hesitant about this one because of the amount of code it affects. I've looked through your code on each iteration but it still feels to me like it complicates the class hierarchy too much, but for the most part the methods are great. In other words, the only thing that worries me is the _implementation_ and the consequences it has. I also have to retract a bit on my suggestion of DataSources - although that solution would allow us to do this, those models **really** belong in the code itself (as a part of FAL).

Recently I began considering the use of Traits to add specific capabilities to a ViewHelper, for example a RenderAssistTrait which contains a `$this->renderChildrenWithVariables(array $variables)` (á la current ViewHelperUtility) and `$this->registerAsArgument` method. We could use such a Trait or set of Traits that contain `$this->registerSlideArguments()` methods (á la `registerUniversalTagAttributes`) and `getSlideRecords`. Such a Trait could then be changed in a subclassed ViewHelper to use another resource type for the sliding, but would re-use the same `registerSlideArguments` method. Maybe that can even be done by just implementing two (or more) Traits.

This could also be further extended to place FAL capabilities in a Trait that *any* ViewHelper can implement and call a simple `register*****Arguments` method from `initializeArguments` and gain access to methods like `getResources` inside the ViewHelper.

Traits would be the ideal way to avoid code duplication without affecting the actual class hierarchy (we can even loosen it up a bit more which would be great also for transparency), but it has a **heap** of angles that should be considered. I think after all this work you have a great perspective on this so I would like to invite you to discuss this on IRC so we can get into all the little details and maybe create a working proof of concept together. You can use the [support chat plugin](https://fluidtypo3.org/community/support-chat.html) if you don't use an IRC client. The best time to catch me is from 19:00 to around midnight, or on the weekend - and you're welcome to message me directly.

NamelessCoder:
Almost forgot to mention: it looks like most of the method bodies you already wrote could be used for the suggested Traits, so hopefully I haven't wasted your time with this re-planning. I very much appreciate the time you spent on this!

cedricziel:
Hmm, I'm also concerned about side effects this may introduce. Can you please rebase it, so we can test? While you're at it, please correct the codestyle errors; thank you!

pkerling:
@cedricziel: It has become clear that it will not be implemented/merged as it is now, so a rebase does not make sense. I will talk with NamelessCoder about the direction of this feature and implement a different solution then, but it might take some time.

pkerling:
I've scrambled together at a9be9e6af86f3519b18a2093b71c94adc1ef8b3b a rough draft to use as a starting point for discussion - sorry that it's taken so long. It's nothing exiciting though, only `AbstractSlideViewHelper` from the first draft implemented as `SlideViewHelperTrait`. I'll try to contact you in IRC then about the FAL stuff.

PR #755, opened 2015-01-18
[TASK] CGL - Define classes, params, use imports

xf-:


NamelessCoder:
Some imports missing line break after license header - I assume your script can also fix these when we run it on VHS ;)

PR #757, opened 2015-01-24
[FEATURE] iterator.merge: Allow use in a chain

neufeind:
Allow use like:
{abc -> v:iterator.merge(a: def)}

This makes it easier to join multiple arrays
in a row and also to workaround some cases
where the Fluid-parser might otherwise
have a hard time.

### DIFF
--- a/Classes/ViewHelpers/Iterator/MergeViewHelper.php
+++ b/Classes/ViewHelpers/Iterator/MergeViewHelper.php
@@ -24,13 +24,15 @@ class MergeViewHelper extends AbstractViewHelper {
 	/**
 	 * Merges arrays/Traversables $a and $b into an array
 	 *
-	 * @param mixed $a First array/Traversable
 	 * @param mixed $b Second array/Traversable
+	 * @param mixed $a First array/Traversable - if not set, the ViewHelper can be in a chain (inline-notation)
 	 * @param boolean $useKeys If TRUE, comparison is done while also observing (and merging) the keys used in each array
 	 * @return array
 	 */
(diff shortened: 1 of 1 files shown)


NamelessCoder:
#### This is an automated comment based on an automated formal commit review.

Your pull request contains formal errors. Comments have been assigned to each commit in your pull request - please review and adjust. Feel free to ask for help if you need it!

NamelessCoder:
Suggested method body:

```php
/**
	 * Merges arrays/Traversables $a and $b into an array
	 *
	 * @param mixed $b Second array/Traversable
	 * @param mixed $a First array/Traversable
	 * @param boolean $useKeys If TRUE, comparison is done while also observing (and merging) the keys used in each array
	 * @return array
	 */
	public function render($b, $a = NULL, $useKeys = TRUE) {
		if (NULL === $a) {
			$a = ViewHelperUtility::arrayFromArrayOrTraversableOrCSV($a, $useKeys);
		}
		$b = ViewHelperUtility::arrayFromArrayOrTraversableOrCSV($b, $useKeys);
		$merged = GeneralUtility::array_merge_recursive_overrule($a, $b);
		return $merged;
	}
```

* Matches usages of v:math.* arguments a/b.
* Gets by with a single use of recursive array merging
* Is somewhat easier to follow, logic-wise

neufeind:
Pushed new version like you suggested. I'm unsure if we should demand a or b to be set, so the original idea was to still allow to be use both :-) But I'm also fine with the current solution.
Added adjusted commit-msg. Be sure the correct one is used upon merge. (I'm not sure github picks the right one.)

PR #760, opened 2015-01-28
[BUGFIX] Use original page uid for submenu rendering

bjo3rnf:
When ``useShortcutUid="true"`` submenu items could not be fetched because of the overridden UID in the page record array.

### DIFF
--- a/Classes/ViewHelpers/Page/Menu/AbstractMenuViewHelper.php
+++ b/Classes/ViewHelpers/Page/Menu/AbstractMenuViewHelper.php
@@ -380,7 +380,7 @@ abstract class AbstractMenuViewHelper extends AbstractTagBasedViewHelper {
 	 */
 	protected function getMenuItemEntry($page, $rootLine, array $parentPage = NULL) {
 		$getLL = $GLOBALS['TSFE']->sys_language_uid;
-		$pageUid = $originalPageUid = $submenuPid = $page['uid'];
+		$pageUid = $page['originalPageUid'] = $page['uid'];
 		$targetPage = NULL;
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Page/Menu/AbstractMenuViewHelper.php
```diff
@@ -388,9 +388,7 @@ protected function getMenuItemEntry($page, $rootLine, array $parentPage = NULL)
 		}
 		if (PageRepository::DOKTYPE_MOUNTPOINT === $doktype) {
 			$mountInfo = $GLOBALS['TSFE']->sys_page->getMountPointInfo($page['uid'], $page);
-			$mountedPageUid = $mountInfo['mount_pid'];
```
NamelessCoder: I don't doubt that these lines are necessary, but could you explain what the operations here are intended to do? Is it safe to remove `$submenuPid` and should we be using the `originalPageUid` as argument for `getMountPointInfo`...?

Basically

bjo3rnf: I did this mainly to reduce the number of variables which contained the same values anyway in some cases. Fx. ``$submenuPid`` should always equal ``$originalPageUid`` and is the original target of this PR. IMHO when using ``$useShortcutUid`` this should only affect the link of the page to avoid a redirect and not determination of the link's state or context.

bjo3rnf:
There's one more thing to fix in a similar fashion plus I forgot to create a topic branch first so I'll close this and reopen a fresh one.

PR #768, opened 2015-02-11
[BUGFIX] Use original page uid for submenu rendering and correct page overlay

bjo3rnf:
This patch fixes:

- when ``useShortcutUid="true"`` submenu items can not be fetched because of the overridden UID in the page record array
- shortcuts in a multilanguage setup are always translated with the target page overlay


NamelessCoder:
The next ten seconds of that gif loop is absolute carnage ^^

bjo3rnf:
"The following program contains scenes of violence which may not be suitable for younger children. Parental supervision is advised"

PR #772, opened 2015-02-20
Cleanups

cedricziel:


cedricziel:
Ahh. Forgot to rebase. Will do later.

cedricziel:
just remade it. There's a lot of inconsistencies I tried to address.
Though we're dynamically typed, i think especially announced return types should match.

NamelessCoder:
* Needs rebase
* The `Tx_Extbase_DomainObject_DomainObjectInterface` usage can be switched to the new class name
* When the class is global, just reference it using `\RuntimeException` and so on - no need to import those.
* And what @bjo3rnf said :)

cedricziel:
Crap-a-chino. Once again merge conflicts. Stop working ^^

PR #773, opened 2015-02-25
[TASK] Add wmode param to YoutubeViewHelper

hhoechtl:
In IE10, IE11 the youtube player ignores its z-index.
You need to set this parameter to make the player behave like it’s
expected.

See
http://stackoverflow.com/questions/4050999/youtube-iframe-wmode-issue
for further details.


Classes/ViewHelpers/Media/YoutubeViewHelper.php
```diff
@@ -60,6 +60,7 @@ public function initializeArguments() {
 		$this->registerArgument('end', 'integer', 'Stop playing after seconds.', FALSE);
 		$this->registerArgument('lightTheme', 'boolean', 'Use the YouTube player\'s light theme.', FALSE, FALSE);
 		$this->registerArgument('videoQuality', 'string', 'Set the YouTube player\'s video quality (hd1080,hd720,highres,large,medium,small).', FALSE);
+		$this->registerArgument('wmode', 'string', 'Set the Window-Mode of the YouTube player (transparent,opaque). This is necessary for z-index handling in IE10/11.', FALSE);
```
NamelessCoder: Should have been named `windowMode`. Not a single one of the other parameters use a contracted naming so this attribute shouldn't either.

PR #775, opened 2015-02-27
[BUGFIX] Arguments L viewhelper not registered

Madii:
The arguments of the `l` viewhelper were not registered so i added them.

### DIFF
--- a/Classes/ViewHelpers/LViewHelper.php
+++ b/Classes/ViewHelpers/LViewHelper.php
@@ -33,6 +33,18 @@ use TYPO3\CMS\Extbase\Utility\LocalizationUtility;
 class LViewHelper extends TranslateViewHelper {
 
 	/**
+	 * Initialize arguments
+	 */
+	public function initializeArguments() {
+		$this->registerArgument('arguments', 'array', 'Translation Key', FALSE, NULL);
+		$this->registerArgument('defualt', 'string', 'Translation Key', FALSE, NULL);
+		$this->registerArgument('extensionName', 'string', 'Translation Key', FALSE, NULL);
+		$this->registerArgument('khtmlEscape', 'boolean', 'Translation Key');
+		$this->registerArgument('id', 'string', 'Translation Key', FALSE, NULL);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
#### This is an automated comment based on an automated formal commit review.

Your pull request contains formal errors. Comments have been assigned to each commit in your pull request - please review and adjust. Feel free to ask for help if you need it!

NamelessCoder:
#### This is an automated comment based on an automated formal commit review.

Your pull request contains formal errors. Comments have been assigned to each commit in your pull request - please review and adjust. Feel free to ask for help if you need it!

NamelessCoder:
The reason for this problem is that apparently, in TYPO3 7.1 those arguments are now arguments for the `render()` method. Overriding the `initializeArguments` method in `v:l` makes, sense, but:

* Your commit sets every argument description and default value to the same value. Copy the descriptions and values from https://github.com/TYPO3/TYPO3.CMS/blob/TYPO3_6-2-10/typo3/sysext/fluid/Classes/ViewHelpers/TranslateViewHelper.php#L74.
* After this change we no longer use any part of the original TranslateViewHelper from Fluid - which means the parent class of `v:l` should be changed to the standard `AbstractViewHelper` class.

Your commits also break our commit rules. The first one is okay, but the following ones should have been made as amends, not new commits. We recommend you create a new branch for such changes so that you can do such changes without worrying about the state of your `development` branch. For the full information, see https://fluidtypo3.org/documentation/contributing/contribution-guide.html.

The TL;DR: clone the repository to your local machine. Run `composer install` and use `./vendor/bin/make` to install hooks for Git which run tests and validate the commit style. Use `./vendor/bin/runtests` to verify the code works - and `./vendor/bin/checkstyle` to verify that the coding style is okay.

Madii:
ok i have done it. But i  wasn't able to run the unittests. Got an composer error when it tried to install typo3 cms. Rest is fine. I will commit the new code.

cedricziel:
Rebased and squashed for you in #779

PR #784, opened 2015-03-09
[BUGFIX] Implode by linefeed

### LINKED ISSUE #783: [BUG] Array output in ContentRender
https://github.com/FluidTYPO3/vhs/commit/026b45b5a8ad79322fb5510710263a3bd652bb3d

Since this line is missing, you get an array instead of the content-element. Bug or Feature? 

$content = implode(LF, $content);

cedricziel:
The content variable still needs to be imploded as otherwise this
was GetViewHelper behaviour

Resolves #783


cedricziel:
I agree. Fix was proposed.

PR #786, opened 2015-03-14
[BUGFIX] Remove redundant slash in URL

jkphl:
`TYPO3_SITE_URL` always comes with a trailing slash ...


NamelessCoder:
Thanks @jkphl - just one Q: are you sure that this parameter is affected by neither config.absRefPrefix nor config.baseURL? We'd need to confirm this - if the value can't be trusted (e.g. comes directly from user input in TS), then the safe solution would be to `rtrim` the last slash off the env variable and `ltrim` it off the `src` and then concatenate those with a single slash.

jkphl:
@NamelessCoder You're right, that'd be the fool-proof way. We should handle it like this anyway. At present, however, `TYPO3_SITE_URL` always [comes with a trailing slash](https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L3433) under normal circumstances, so I think simplifying it by only `ltrim`ming the `$src` is pretty safe as well.

NamelessCoder:
Thanks for the verification - yep, this ENV is perfectly trustworthy :)

PR #788, opened 2015-03-15
[TASK] Remove duplicate code

monosize:
usingsortArrayto sort $objectStorage


Classes/ViewHelpers/Iterator/SortViewHelper.php
```diff
@@ -73,8 +73,8 @@ public function initializeArguments() {
 	 *
 	 * Returns the same type as $subject. Ignores NULL values which would be
 	 * OK to use in an f:for (empty loop as result)
-	 *
```
NamelessCoder: Letting slide because it's cosmetics, but please don't remove existing whitespace in phpdoc comments. CGL calls for one empty space after description before param/return/throws and no space only when there is no description.

Issue #790, opened 2015-03-15
v:page.header.meta renders tag with contents


aksenovaa:
If I use <v:page.header.meta name="viewport" content="width=device-width, initial-scale=1.0"/> or
{v:page.header.meta(name: 'viewport', content: 'width=device-width, initial-scale=1.0')}

Then the output to the page I get:
<meta name="viewport" content="width=device-width, initial-scale=1.0">width=device-width, initial-scale=1.0</meta>
instead of
<meta name="viewport" content="width=device-width, initial-scale=1.0" ./>

TYPO3 v 7.1

typo3ua:
I have the same. I update VHS to version 2.3.1 ... TYPO3 v 6.2.10
My code:
~~~
<v:page.header.meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no" />
~~~
Output:
~~~
<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no">width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no</meta>
~~~

rabe69:
Can confirm this bug.
The problem seems to come from the changes in MetaViewHelper.php
If I change line 57 from this:
$GLOBALS['TSFE']->getPageRenderer()->addMetaTag($this->renderTag($this->tagName, $this->arguments['content']));
to this:
$GLOBALS['TSFE']->getPageRenderer()->addMetaTag($this->renderTag($this->tagName));
The output is correct.

mhmli:
Same here. Same problem, same solution.

PR #791, opened 2015-03-16
[BUGFIX] Correctly output meta tags

tantegerda1:
Html meta tags do not have a "content", they might have a "content" _attribute_.

See http://www.w3.org/wiki/HTML/Elements/meta


tantegerda1:
fixes #790

NamelessCoder:
Thanks for the patch Ludwig! Just one thing: we need the `['content' => $this->arguments['content']]` to be written as `array('content' => $this->arguments['content'])` - until such a time we are ready to convert all current usages of `array()` to `[]`.

NamelessCoder:
Thanks - very much appreciated!

Manually picked and squashed as https://github.com/FluidTYPO3/vhs/commit/a460c9f9b8fd9901c2c006ddaadc06e21cfbd531, will be in VHS 2.3.2.

PR #793, opened 2015-03-18
Development

maikp:
[BUGFIX] assertFrontendUserLoggedIn returns FALSE if no frontendUser is given but the currentFrontentUser is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@ VHS: Fluid ViewHelpers
 > Collection of general purpose ViewHelpers usable in the Fluid templating engine
 > that's bundled with the TYPO3 CMS.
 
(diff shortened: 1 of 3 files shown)


xf-:
i think there is something wrong with your commit.

maikp:
Yes, this schould be ignored.

I tried to fix #794 with the following code:

	public function assertFrontendUserLoggedIn(FrontendUser $frontendUser = NULL) {
		$currentFrontendUser = $this->getCurrentFrontendUser();
		if (TRUE === $currentFrontendUser instanceof FrontendUser) {
			if (TRUE === $frontendUser instanceof FrontendUser) {
				if (FALSE === ($currentFrontendUser->getUid() === $frontendUser->getUid())) {
					return FALSE;
				}
			}
			return TRUE;
		}
		return FALSE;
	}


But I can't upload this change. It would be nice if someone else con do this change in 
vhs/Classes/ViewHelpers/Security/AbstractSecurityViewHelper.php

Issue #794, opened 2015-03-18
assertFrontendUserLoggedIn always returns FALSE if no frontentuser given


maikp:
assertFrontendUserLoggedIn returns FALSE if no frontendUser is given but the currentFrontentUser is set.

mrk131324:
Here is a solution:

Classes/ViewHelpers/Security/AbstractSecurityViewHelper.php:130:

    /**
     * Returns TRUE only if a FrontendUser is currently logged in. Use argument
     * to return TRUE only if the FrontendUser logged in must be that specific user.
     *
     * @param FrontendUser $frontendUser
     * @return boolean
     * @api
     */
    public function assertFrontendUserLoggedIn(FrontendUser $frontendUser = NULL) {
        $currentFrontendUser = $this->getCurrentFrontendUser();
		if (FALSE === $frontendUser instanceof FrontendUser) {
			$frontendUser = $this->getCurrentFrontendUser();
		}
		if (TRUE === $frontendUser instanceof FrontendUser && TRUE === $currentFrontendUser instanceof FrontendUser) {
			if ($currentFrontendUser->getUid() === $frontendUser->getUid()) {
				return TRUE;
			} else {
				return FALSE;
			}
		}

		return FALSE;
	}

maikp:
|Yes, but I think this is simpler and much easier to read:


public function assertFrontendUserLoggedIn(FrontendUser $frontendUser = NULL) {
     $currentFrontendUser = $this->getCurrentFrontendUser();
     if (TRUE === $currentFrontendUser instanceof FrontendUser) {
         if (TRUE === $frontendUser instanceof FrontendUser) {
             return $currentFrontendUser->getUid() === $frontendUser->getUid();
         }
         return TRUE;
     }
     return FALSE;
}|

Am 24.03.2015 um 20:35 schrieb Martin R. Krause:

bjo3rnf:
Hi @maikp,

mind recreating your PR? In case you need some help don't hesitate to ask. Also please keep in mind that you'll have to change the according unit test as well as this changes the original behavior.

Cheers
Björn

loco8878:
the 'soloution' does not work. Method evaluateArguments call Method assertFrontendUserLoggedIn without arguments. You get always false in assertFrontendUserLoggedIn Method.

PR #796, opened 2015-03-18
[BUGFIX] assertFrontendUserLoggedIn returns FALSE

maikp:
if no frontendUser is given but the currentFrontentUser is set.


maikp:
Fixes #794

NamelessCoder:
Please recreate pull request - current PR is using upstream `development` as source and upstream `staging` as target. Needs to use your fork's PR branch as source and upstream's `development` branch as target.

PR #798, opened 2015-03-23
[FEATURE] Add attribute "property" to Page\Header\MetaViewHelper

monosize:
missing attribute "property" for e.g. to set open graph metadata in page header.


NamelessCoder:
Not sure I understand/agree with this one? The `<meta>` tag has no `property` attribute and the newly added ViewHelper argument is never used...?

monosize:
I see, it's not an official attribute.

But it's used for e.g. open graph metadata.
<meta property="og:title" content="Title" />

more Information here: http://moz.com/blog/meta-data-templates-123

NamelessCoder:
Aaaah like so - well, that makes perfect sense to me. Screw the specs ^^.

PR #809, opened 2015-04-08
Update AbstractSecurityViewHelper.php

### LINKED ISSUE #794: assertFrontendUserLoggedIn always returns FALSE if no frontentuser given

loco8878:
fixes #794


bjo3rnf:
Manually picked in 3241fc374866fbedd7d6232eb3aed36857763dde. Thanks for contributing!

PR #813, opened 2015-04-14
[FEATURE] New viewhelper /Resource/Folder

monosize:
ViewHelper to handle FAL folders, fetch all files of the given folders.
An additional filter on file extensions can be used.
The identifier must be in the format '[fileStorageId]:/folder/subfolder/'
Example identifier: 'file:1:/myFolder/' or '1:/myFolder/'

Reorganize ```Classes/Resource/AbstractResourceViewHelper.php```. Move resource based code to Trait ```Classes/Traits/ResourceViewHelperTrait.php```

Added the method ```convertValueOrGetArgumentAsArray```to ```Classes/Traits/Trait ArrayConsumingViewHelperTrait```, formerly it was the method ```arrayForMixedArgument```from  ```Classes/Resource/AbstractResourceViewHelper.php```

### DIFF
--- a/Classes/Traits/ResourceViewHelperTrait.php
+++ b/Classes/Traits/ResourceViewHelperTrait.php
@@ -1,0 +1,198 @@
+<?php
+namespace FluidTYPO3\Vhs\Traits;
+
+/*
+ * This file is part of the FluidTYPO3/Vhs project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use TYPO3\CMS\Core\Resource\Filter\FileExtensionFilter;
+use TYPO3\CMS\Core\Resource\Folder;
+use TYPO3\CMS\Core\Utility\ArrayUtility;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Persistence\QueryResultInterface;
+use TYPO3\CMS\Fluid\Core\ViewHelper\Exception;
+use FluidTYPO3\Vhs\Utility\ResourceUtility;
+
+
+/**
+ * Class ResourceViewHelperTrait
+ *
+ * Trait implemented by ViewHelpers that operate with
+ * File and Folder Resources
+ *
+ * Contains the following main responsibilities:
+ *
+ * - get files of given identifiers or categories
+ * - fetch files of folders, given by FAL identifier.
+ *   With the ability to filter the files on given file extensions.
+ */
+trait ResourceViewHelperTrait {
+
+	use ArrayConsumingViewHelperTrait;
+
+	/**
+	 *
+	 * @var FileExtensionFilter
+	 */
+	protected $filter = NULL;
+
+	/**
+	 * Returns the files
+	 *
+	 * @param boolean $onlyProperties
+	 * @param mixed $identifier
+	 * @param mixed $categories
+	 * @return array|NULL
+	 */
+	public function getFiles($onlyProperties = FALSE, $identifier = NULL, $categories = NULL) {
+		$identifier = $this->convertValueOrGetArgumentAsArray($identifier, 'identifier');
+		$categories = $this->convertValueOrGetArgumentAsArray($categories, 'categories');
+		$treatIdAsUid = (boolean) $this->arguments['treatIdAsUid'];
+		$treatIdAsReference = (boolean) $this->arguments['treatIdAsReference'];
+
+		if (TRUE === $treatIdAsUid && TRUE === $treatIdAsReference) {
+			throw new \RuntimeException('The arguments "treatIdAsUid" and "treatIdAsReference" may not both be TRUE.', 1384604695);
+		}
+
+		if (TRUE === empty($identifier) && TRUE === empty($categories)) {
+			return NULL;
+		}
+
+		$files = array();
+		$resourceFactory = GeneralUtility::makeInstance('TYPO3\\CMS\\Core\\Resource\\ResourceFactory');
+
+		if (FALSE === empty($categories)) {
+			$sqlCategories = implode(',', $GLOBALS['TYPO3_DB']->fullQuoteArray($categories, 'sys_category_record_mm'));
+			$res = $GLOBALS['TYPO3_DB']->exec_SELECTquery('uid_foreign', 'sys_category_record_mm', 'tablenames = \'sys_file\' AND uid_local IN (' . $sqlCategories . ')');
+
+			$fileUids = array();
+			while ($row = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($res)) {
+				$fileUids[] = intval($row['uid_foreign']);
+			}
+			$fileUids = array_unique($fileUids);
+
+			if (TRUE === empty($identifier)) {
+				foreach ($fileUids as $fileUid) {
+					try {
+						$file = $resourceFactory->getFileObject($fileUid);
+
+						if (TRUE === $onlyProperties) {
+							$file = ResourceUtility::getFileArray($file);
+						}
+
+						$files[] = $file;
+					} catch (\Exception $e) {
+						continue;
+					}
+				}
+
+				return $files;
+			}
+		}
+
+		foreach ($identifier as $i) {
+			try {
+				if (TRUE === $treatIdAsUid) {
+					$file = $resourceFactory->getFileObject(intval($i));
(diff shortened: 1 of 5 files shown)


Classes/ViewHelpers/Resource/FolderViewHelper.php
```diff
@@ -0,0 +1,62 @@
...
+
+		$this->registerAsArgument();
+	}
+
+	/**
+	 * @return mixed
+	 */
+	public function render() {
+		$files = $this->getFilesOfFolders(TRUE);
+		if (1 === count($files)) {
```
NamelessCoder: This magic behavior should be avoided. It makes the return value unpredictable. The class header could be fitted with a small example how to use `-> v:iterator.first()` to retrieve the first record when using the ViewHelper with and without the `as` argument.

Classes/Traits/ResourceViewHelperTrait.php
```diff
@@ -0,0 +1,198 @@
...
+				if (NULL !== $this->filter) {
+					$folder->setFileAndFolderNameFilters($filter);
+				}
+
+				if (TRUE === $onlyProperties) {
+					$files = array_merge($files, self::getFilesOfFolderAsArray($folder, $start, $numberOfItems, $recursive));
+				} else {
+					$files = array_merge($files, $folder->getFiles($start, $numberOfItems, Folder::FILTER_MODE_USE_OWN_AND_STORAGE_FILTERS, $recursive));
+				}
+			} catch (\Exception $e) {
```
NamelessCoder: At the very least, this error should be logged. I suggest re-throwing it as a ViewHelper/Exception.

Classes/Traits/ResourceViewHelperTrait.php
```diff
@@ -0,0 +1,198 @@
...
+ */
+trait ResourceViewHelperTrait {
+
+	use ArrayConsumingViewHelperTrait;
+
+	/**
+	 *
+	 * @var FileExtensionFilter
+	 */
+	protected $filter = NULL;
```
NamelessCoder: Please pass this as method argument rather than contain it in a property that's mutated.

I'm not too sure about this Trait using another Trait. It's not necessarily true that any resource ViewHelper should also be able to process arrays. (see below about convertValueOrGetArgumentAsArray).

Classes/Traits/ResourceViewHelperTrait.php
```diff
@@ -0,0 +1,198 @@
...
+	/**
+	 * Returns the files
+	 *
+	 * @param boolean $onlyProperties
+	 * @param mixed $identifier
+	 * @param mixed $categories
+	 * @return array|NULL
+	 */
+	public function getFiles($onlyProperties = FALSE, $identifier = NULL, $categories = NULL) {
+		$identifier = $this->convertValueOrGetArgumentAsArray($identifier, 'identifier');
```
NamelessCoder: This value should be extracted from the ViewHelper, not the Trait. In addition, the name of the expected arguments are hardcoded here but will only fit the FolderViewHelper. I suggest:

* pass the `$identifiers` and `$categories` as arguments
* similar to the `registerAsArgument`, create a `registerResourceArguments` which can register universal and homogenous named arguments for resource-related ViewHelpers (treatIdAsReference etc.)

Classes/Traits/ResourceViewHelperTrait.php
```diff
@@ -0,0 +1,198 @@
...
+/*
+ * This file is part of the FluidTYPO3/Vhs project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use TYPO3\CMS\Core\Resource\Filter\FileExtensionFilter;
+use TYPO3\CMS\Core\Resource\Folder;
+use TYPO3\CMS\Core\Utility\ArrayUtility;
```
NamelessCoder: This class, very unfortunately, cannot be trusted. However: the features depended on from this class should already be present in the ArrayConsumingViewHelperTrait. Decoupling these two Traits should remove the need for this import.

Classes/ViewHelpers/Resource/FolderViewHelper.php
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * @return void
+	 * @api
+	 */
+	public function initializeArguments() {
+		parent::initializeArguments();
+		$this->registerArgument('identifier', 'mixed', 'The FAL combined identifiers (either CSV, array or implementing Traversable).', FALSE, NULL);
+		$this->registerArgument('recursive', 'boolean', 'If TRUE,then all files are fetched recursivly.', FALSE, FALSE);
+		$this->registerArgument('start', 'int', 'The item to start at.', FALSE, 0);
+		$this->registerArgument('numberOfItems', 'int', 'The number of items to return', FALSE, 0);
```
NamelessCoder: Probably needs to add `, zero for all items` to the description.

Classes/Traits/ResourceViewHelperTrait.php
```diff
@@ -0,0 +1,198 @@
...
+	 * Get all Files from the $folder and convert each file object to an array.
+	 *
+	 * @param Folder $folder
+	 * @param int    $start
+	 * @param int    $numberOfItems
+	 * @param bool   $recursive
+	 *
+	 * @return array
+	 */
+	public static function getFilesOfFolderAsArray(Folder $folder, $start = 0, $numberOfItems = 0, $recursive = FALSE) {
```
NamelessCoder: Definitely shouldn't be static ;)

Classes/Traits/ResourceViewHelperTrait.php
```diff
@@ -0,0 +1,198 @@
...
+	 * Returns the files of the folders
+	 *
+	 * @param boolean $onlyProperties
+	 * @param mixed   $identifier
+	 * @param mixed   $recursive
+	 *
+	 * @return array|NULL
+	 */
+	public function getFilesOfFolders($onlyProperties = FALSE, $identifier = NULL, $recursive = NULL) {
+		$identifier = $this->convertValueOrGetArgumentAsArray($identifier, 'identifier');
```
NamelessCoder: The arguments extracted here have to be added as method arguments instead; see above.

NamelessCoder:
In addition to the inline comments, a few overall notes:

Top priority:

* Avoid catching Exceptions; always catch a subclass. It is impossible to test code that catches root Exceptions.

Also important:

* I would definitely split the category- and identifier-based retrieving into separate methods.
* Traits shouldn't inherit from Traits or expect other Traits to be implemented. Instead, the methods that a Trait adds should (just an example) expect already converted/retrieved values **unless** the argument was added by the Trait itself (like the `as` argument is).
* All occurrences of `int` shall be `integer`, all occurrences of `bool` shall be `boolean` - in annotations as well as ViewHelper argument descriptions.
* Argument descriptions should not end with a dot (they are expressed in a table).
* No need to `@api` tag; remove that.
* `getFiles` should always return an array, a `NULL` return should not be possible. The return value and type should be completely predictable. Note: `f:if` will also render `else` if given an empty array as `condition`.
* Fully qualified class names should be imported when used in annotations (there's one for an Exception thrown).

Should also be done as part of this, since you're touching the code anyway:

* treatIdAsUid should be the behavior when treatIdAsReference is FALSE; no need to explicitly allow both arguments. Maybe now is a good time to drop that argument and adjust the behavior.

But I think we can improve the flexibility of this quite a lot with a few changes to the structure. I suggest:

* Rather than just construct a filter, I would add an argument that all resource ViewHelpers can use and which accepts an instance of a filter. This will allow us to create filters using other ViewHelpers; if no filter is given a default one can be created.
* Drop the `onlyProperties` argument in favor of an additional ViewHelper that can load that array, e.g. `v:resource.info` or similar (so it matches `v:content.info` and `v:page.info`).
* Drop the `filterExtensions` argument in favor of a filtering ViewHelper that can be used inline as argument for the previously suggested `filter` argument on resource ViewHelpers.

NamelessCoder:
Needs a little rebasing after merging https://github.com/FluidTYPO3/vhs/pull/679 - and adjustment to review comments. The suggested structure with filters does *not* need to be implemented but the comments and in particular Trait interdependency must be addressed before we merge.

cedricziel:
I'll close this one, as there's no activity since september. If you want to reopen-no problem, just amend the code.

Thank you!

grimmcreative:
Would be so nice if this gets realized.

PR #815, opened 2015-04-16
[BUGFIX] Do not group by uid

andreasfernandez:
Since uid is a unique field, it's not required to group it.
The query fails on other DBMS than MySQL due to the grouping.


cedricziel:
Thx a bunch Andreas!

PR #825, opened 2015-04-27
Allow record sliding in page.resources.fal with traits

pkerling:
Continuation from #679 - now with 100% more traits.

### DIFF
--- a/Classes/Traits/SlideViewHelperTrait.php
+++ b/Classes/Traits/SlideViewHelperTrait.php
@@ -1,0 +1,137 @@
+<?php
+namespace FluidTYPO3\Vhs\Traits;
+
+/*
+ * This file is part of the FluidTYPO3/Vhs project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+/**
+ * Class SlideViewHelperTrait
+ *
+ * Trait implemented by ViewHelpers that wants some kind of
+ * records to be optionally looked up and/or collected from
+ * the current page and pages up the rootline. ViewHelpers must
+ * implement the getSlideRecordsFromPage method which looks up
+ * resources for a single page.
+ *
+ * Has the following main responsibilities:
+ * - register arguments common for sliding
+ * - method to get records with sliding according
+ *   to the ViewHelper arguments
+ */
+trait SlideViewHelperTrait {
+
+	/**
+	 * Default initialisation of arguments - will be used
+	 * if the implementing ViewHelper does not itself define
+	 * this method.
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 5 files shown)


Classes/Traits/SlideViewHelperTrait.php
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	protected function registerSlideArguments() {
+		$this->registerArgument('limit', 'integer', 'Optional limit to the total number of records to render');
+		$this->registerArgument('slide', 'integer', 'Enables Record Sliding - amount of levels which shall get walked up the rootline, including the current page. For infinite sliding (till the rootpage) set to -1. Only the first PID which has at minimum one record is used', FALSE, 0);
+		$this->registerArgument('slideCollect', 'integer', 'If TRUE, content is collected up the root line. If FALSE, only the first PID which has content is used. If greater than zero, this value overrides $slide.', FALSE, 0);
+		$this->registerArgument('slideCollectReverse', 'boolean', 'Normally when collecting records the elements from the actual page get shown on the top and those from the parent pages below those. You can invert this behaviour (actual page elements at bottom) by setting this flag))', FALSE, 0);
+	}
+
+	/**
+	 * @return \TYPO3\CMS\Extbase\Object\ObjectManagerInterface
```
NamelessCoder: `@return FluidTYPO3\Vhs\Service\PageSelectService`

NamelessCoder:
Damn this looks good. Strike that - damn, this looks **really** good! I think there's only nit-picking for style and meta-doc and then adjusting the unit tests (and for 1,000 bonus points, a basic test or two to cover the behavior of this trait!).

pkerling:
I've taken care of the meta-doc miss you pointed out. As far as I can see, none of the existing unit tests need to be changed since they don't do much anyway.
I've also looked into making unit tests for the slide behaviour, but I have zero experience with PHPUnit/TYPO3 unit testing and am completely lost on how to intercept the TYPO3_DB SQL queries and provide static fixtures as results.

xf-:
Hi,
we need to discuss this. This would mean we would drop 6.2 requirements, because traits are introduced in PHP 5.4 and TYPO3 6.2 supports 5.3.7+ and Debian squeeze LTS is until 2016-02.

Personally i would love to drop 5.3 and also use standard array syntax with [] and no extending with [] and definition array().

pkerling:
I don't think there's any change from the status quo. `vhs` already requires PHP 5.4 since traits were introduced in other ViewHelpers about 5 months ago by NamelessCoder.

xf-:
@pkerling ok, that means we drop 5.3 :)

pkerling:
Just had a look at ext_emconf.php, 5.3 is gone since version 2.3 (see commit d1b732d)

xf-:
OK thx :)

NamelessCoder:
Thank you so much again, @pkerling - not least for your patience with this one :)

PR #831, opened 2015-05-15
[FEATURE] Add nonAlphanumeric option to format.eleminate viewhelper

BenjaminBeck:


NamelessCoder:
Suggest using `<v:format.regularExpression pattern="/[^a-z0-9]*/" replacement="">...</v:format.regularExpression>` which also has more control over which characters to preserve. The problem with using `v:format.eliminate` for this is that we would either have to decide for all users what is to be considered alpha-numeric (this differs from locale to locale) or we would have to allow specifying the regular expression - which would duplicate functionality from `v:format.regularExpression`, making it smarter to simply use that in the first place.

Makes sense?

BenjaminBeck:
Yes, i had a wrong definition of alphanumeric in my head ..

PR #832, opened 2015-05-20
[BUGFIX] AlternateViewHelper should include a link to current language

zauni:
The hreflang link Tag should also include the current language. Right now it's deleted, but this is wrong behaviour.

From the [google docs](https://support.google.com/webmasters/answer/189077?hl=en): If you have multiple language versions of a URL, each language page must identify all language versions, **including itself**.


cedricziel:

DenisMir:
It would be great if this fix would make it into the release version.

PR #859, opened 2015-07-10
[BUGFIX] Force closing tag and render comment

### LINKED ISSUE #858: plugin.tx_vhs.settings.asset leads to corrupt style element
This took my literally hours.
I had following TS:
```
plugin.tx_vhs.settings.asset {
	mainCss {
		path = EXT:speciality/Resources/Public/Build/StyleSheets/site.min.css
		group = css
		standalone = 1
		rewrite = 0
		type = css
	}
}
```
The Problem was the extension does not exist, because I renamed it. The real problem was, that vhs was generating ```<style type="text/css" />``` which is invalid and break the whole markup.
Please prevent that such an element is generated, same for ```<script />```. If we have an empty style use ```<style type="text/css"></style>```

bjo3rnf:
Fixes #858


xf-:
+1 look good

cedricziel:
Houston we have a merge!

PR #868, opened 2015-07-23
[FEATURE] For FileReference automatically other parameters correctly

neufeind:
So far you needed to use src="{yourmedia.uid}" and treatIdAsReference="1".
Now if you supply a FileReference this will be set automatically.

### DIFF
--- a/Classes/ViewHelpers/Media/Image/AbstractImageInfoViewHelper.php
+++ b/Classes/ViewHelpers/Media/Image/AbstractImageInfoViewHelper.php
@@ -41,7 +41,7 @@ abstract class AbstractImageInfoViewHelper extends AbstractViewHelper {
 	 * @api
 	 */
 	public function initializeArguments() {
-		$this->registerArgument('src', 'string', 'Path to or id of the image file to determine info for.', TRUE);
+		$this->registerArgument('src', 'mixed', 'Path to or id of the image file to determine info for. In case a FileReference is supplied, treatIdAsUid and treatIdAsReference will automatically be activated.', TRUE);
 		$this->registerArgument('treatIdAsUid', 'boolean', 'If TRUE, the path argument is treated as a resource uid.', FALSE, FALSE);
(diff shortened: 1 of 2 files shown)


xf-:
CGL - TRUE === is_object
About the effect: I don't know if that that kind of magic is welcome.

NamelessCoder:
We should use the same behavior here as the root ViewHelpers do regarding treatIdAsReference. IMHO it does make sense to detect this and switch behavior, instead of requiring the user to make this decision in every usage - but it is more important that the API is predictable  and homogenous with native resource ViewHelpers. If this happens in `f:image` etc. then we will follow suit immediately after ;)

In any case, the old argument should behave like an *override* so that if you do specify the argument, whether the value is TRUE or FALSE, it gets used instead. That would also preserve backwards compatible behavior.

Sounds reasonable?

neufeind:
For f:image the parameter for providing a filereference is not "src" but "image". Should we add the magic-handling discussed here and add "image" simply as an alias for "src" in the same go?
I was unsure how we should handle the annotation for "src" since the argument is defined int he parent-class. And there it is mentioned that "src" could even be an array ... but it can't really be an array when we're talking about the image-VH, right? Hmm ...

neufeind:
About the failed travis-checks: Is it okay they fail for the 7.2-builds?

xf-:
Is OK - problems with core changes...

bjo3rnf:
I like this patch.

> For f:image the parameter for providing a filereference is not "src" but "image". Should we add the magic-handling discussed here and add "image" simply as an alias for "src" in the same go?

IMHO the magic should happen when 'image' is used and src should stay as is especially if this is the original behavior of f:image.

NamelessCoder:
Please adjust to @bjo3rnf's comments - with which I completely agree :)

NamelessCoder:
Any news about this one, Stefan? If you're short on time just let us know and we will drive it home ;)

NamelessCoder:
Manually picked - using `src` was quite alright, but the order of execution needed to be changed a little bit (to extract `src` *before* checking if it's a reference, to support passing it as tag content).

Issue #829, opened 2015-05-11
'allowMoveToFooter' is deprecated, update to 'movable'


Outdoorsman:
'allowMoveToFooter' is deprecated according to the changlog but it's everyone is still using it because the auto-generated documentation on FluidTYPO3 doesn't reflect that. Please update it so that it's noted correctly on FluidTYPO3 about 'movable'.
https://fluidtypo3.org/viewhelpers/vhs/development/AssetViewHelper.html

DenisMir:
@Outdoorsman But when trying to use `movable` inside a `<v:asset.script ../>` script helper I'm getting `Argument "movable" was not registered.`. So I think the new argument is not registered in the viewhelper itself. If you look right here this is the case: https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Asset/AbstractAssetViewHelper.php

PR #870, opened 2015-07-24
[TASK] Deprecate argument

bjo3rnf:
Addresses #829


Classes/ViewHelpers/Asset/AbstractAssetViewHelper.php
```diff
@@ -356,6 +357,11 @@ public function getAssetSettings() {
 		$settings = $this->getSettings();
 		$assetSettings = $this->arguments;
 		$assetSettings['type'] = $this->getType();
+		//TODO: Remove with deprecated argument
```
xf-: maybe add info like after 6.2 LTS

NamelessCoder: We don't really know exactly when this will be removed - VHS itself should already be compatible with 7.x branches so we are not forced to bump up another major version when next LTS is released. I would keep as is and just plan to remove those when we do the next major version.

PR #875, opened 2015-07-28
[TASK] Add Responsive Images

mneuhaus:
PictureViewHelper is already quite steady. Srcset is still kind of rough around the edges and only added to ```v:media.image``` so far.

### DIFF
--- a/Classes/ViewHelpers/Media/PictureViewHelper.php
+++ b/Classes/ViewHelpers/Media/PictureViewHelper.php
@@ -1,0 +1,93 @@
+<?php
+namespace FluidTYPO3\Vhs\ViewHelpers\Media;
+
+/*
+ * This file is part of the FluidTYPO3/Vhs project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractTagBasedViewHelper;
+use TYPO3\CMS\Fluid\Core\ViewHelper\Exception;
+use TYPO3\CMS\Fluid\Core\ViewHelper\TagBuilder;
+
+/**
+ * Renders a picture element with different images/sources for specific
+ * media breakpoints
+ *
+ * ### Example
+ *
+ *     <v:media.picture src="fileadmin/some-image.png" alt="Some Image">
+ *         <v:media.source media="(min-width: 1200px)" width="500c" height="500c" />
+ *         <v:media.source media="(min-width: 992px)" width="300c" height="300c" />
+ *         <v:media.source media="(min-width: 768px)" width="200c" height="200c" />
+ *         <v:media.source width="80c" height="80c" />
+ *     </v:media.picture>
+ *
+ * ### Browser Support
+ *
+ * To have the widest Browser-Support you should consider using a polyfill like:
+ * http://scottjehl.github.io/picturefill/
+ *
+ * @author Marc Neuhaus <mneuhaus@famelo.com>, famelo
+ * @package Vhs
+ * @subpackage ViewHelpers\Media
+ */
+class PictureViewHelper extends AbstractTagBasedViewHelper {
+	const SCOPE = 'FluidTYPO3\Vhs\ViewHelpers\Media\PictureViewHelper';
+	const SCOPE_VARIABLE_SRC = 'src';
+	const SCOPE_VARIABLE_DEFAULT_SOURCE = 'default-source';
+
+	/**
+	 * name of the tag to be created by this view helper
+	 *
+	 * @var string
+	 * @api
+	 */
+	protected $tagName = 'picture';
+
+	/**
+	 * Initialize arguments.
+	 *
+	 * @return void
+	 * @api
+	 */
+	public function initializeArguments() {
+		parent::initializeArguments();
+		$this->registerArgument('src', 'string', 'Path to the image.', TRUE);
+		$this->registerArgument('alt', 'string', 'Text for the alt tag.', TRUE);
+		$this->registerArgument('title', 'string', 'Text for the alt tag.', FALSE);
+	}
+
+	/**
+	 * Render method
+	 *
+	 * @return string
+	 */
+	public function render() {
+		$src = $this->arguments['src'];
+
+		$this->viewHelperVariableContainer->addOrUpdate(self::SCOPE, self::SCOPE_VARIABLE_SRC, $src);
+		$content = $this->renderChildren();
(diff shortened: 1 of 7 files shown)


Classes/ViewHelpers/Media/PictureViewHelper.php
```diff
@@ -0,0 +1,92 @@
...
+
+		if ('' === $this->arguments['title']) {
+			$defaultImage->addAttribute('title', $this->arguments['alt']);
+		}
+		$content.= $defaultImage->render();
+
+		$this->tag->setContent($content);
+		return $this->tag->render();
+	}
+}
```
xf-: Please use new line

xf-:
Looks like edge supports src-set
http://blogs.windows.com/msedgedev/2015/06/08/introducing-srcset-responsive-images-in-microsoft-edge/

Classes/ViewHelpers/Media/PictureViewHelper.php
mneuhaus: done :)

Classes/ViewHelpers/Resource/AbstractImageViewHelper.php
```diff
@@ -185,4 +185,4 @@ public function preprocessSourceUri($source) {
 		return $source;
 	}
 
-}
+}
```
xf-: newline please

mneuhaus: doh! :( sorry added newline there as well

cedricziel:
Truly a great addition - thank you!

Can you add a test which at least executes a ViewHelper with the new argument?

Cheers,
Cedric

mneuhaus:
@cedricziel like this? :)

NamelessCoder:
Don't merge yet. Looks okay overall but has tremendous potential impact; frontend simulation is removed among other things. Needs *proper* testing as it affects every image-related ViewHelper. Some CGL issues (missing phpdoc etc) too.

mneuhaus:
As i said in irc, i reverted the change in the ```AbstractMediaViewHelper``` regarding ```FrontendSimulation```. 
I do use the ```FrontendSimulationUtility``` instead of the ```AbstractMediaViewHelper``` variants in the ```SourceSetViewHelperTrait``` which seems to work find in FE + BE.

+ i added missing phpdocs in ```SourceSetViewHelperTrait```. 

Any other CGL issues i'm blind about? :)

Classes/ViewHelpers/Media/PictureViewHelper.php
```diff
@@ -0,0 +1,92 @@
...
+	 */
+	public function render() {
+		$src = $this->arguments['src'];
+
+		$this->viewHelperVariableContainer->addOrUpdate(self::SCOPE, self::SCOPE_VARIABLE_SRC, $src);
+		$content = $this->renderChildren();
+		$this->viewHelperVariableContainer->remove(self::SCOPE, self::SCOPE_VARIABLE_SRC);
+
+		if (FALSE === $this->viewHelperVariableContainer->exists(self::SCOPE, self::SCOPE_VARIABLE_DEFAULT_SOURCE)) {
+			throw new \Exception('Please add a source without a media query as a default.', 1438116616);
```
NamelessCoder: Should not throw a root-type `\Exception` - at the very least it should be a `\RuntimeException` since those can be caught in unit tests. In this case I suggest simply throwing a `TYPO3\CMS\Fluid\Core\ViewHelper\Exception` since that results in a friendly string output instead of a stack trace.

Classes/ViewHelpers/Media/PictureViewHelper.php
```diff
@@ -0,0 +1,92 @@
...
+		}
+		$defaultSource = $this->viewHelperVariableContainer->get(self::SCOPE, self::SCOPE_VARIABLE_DEFAULT_SOURCE);
+
+		$defaultImage = new TagBuilder('img');
+		$defaultImage->addAttribute('src', $defaultSource);
+		$defaultImage->addAttribute('alt', $this->arguments['alt']);
+
+		if ('' === $this->arguments['title']) {
```
NamelessCoder: The default value of `$this->arguments['title']` would be NULL. I would either compare `NULL === ...` or simply use `empty()`.

Classes/ViewHelpers/Media/SourceViewHelper.php
```diff
@@ -0,0 +1,144 @@
...
+ * Used in conjuntion with the `v:media.PictureViewHelper`.
+ * Please take a look at the `v:media.PictureViewHelper` documentation for more
+ * information.
+ *
+ * @author Marc Neuhaus <mneuhaus@famelo.com>, famelo
+ * @package Vhs
+ * @subpackage ViewHelpers\Media
+ */
+class SourceViewHelper extends AbstractTagBasedViewHelper {
+	const SCOPE = 'FluidTYPO3\Vhs\ViewHelpers\Media\PictureViewHelper';
```
NamelessCoder: Line before `const`s start, please :)

Classes/ViewHelpers/Media/SourceViewHelper.php
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * name of the tag to be created by this view helper
+	 *
+	 * @var string
+	 * @api
+	 */
+	protected $tagName = 'source';
+
+	/**
+	 * @var \TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer
```
NamelessCoder: FQN can be imported.

Classes/ViewHelpers/Media/SourceViewHelper.php
```diff
@@ -0,0 +1,144 @@
...
+			$quality = MathUtility::forceIntegerInRange($quality, 10, 100, 75);
+			$setup['params'] .= ' -quality ' . $quality;
+		}
+
+		if ('BE' === TYPO3_MODE && '../' === substr($src, 0, 3)) {
+			$src = substr($src, 3);
+		}
+		$result = $this->contentObject->getImgResource($src, $setup);
+
+		if ('BE' === TYPO3_MODE) {
```
NamelessCoder: I believe there is a call to restore the context after running this? If I remembered that right, then please make sure that method is called before the function `return`s.

mneuhaus: could you elaborate? i don't really get what you mean, i do reset ```FrontendSimulationUtility::resetFrontendEnvironment();``` or do you mean, something that static function doesn't do?

NamelessCoder: `FrontendSimulationUtility::resetFrontendEnvironment();` - yep, that's the one! Just fire that before the method returns and we're good, I think.

PR #879, opened 2015-07-31
[BUGFIX] Force overriding with overlay values

bjo3rnf:
Currently overlay values are only used if not empty which leads to unexpected behavior regarding link title generation in menus. A page may provide values for nav_title *and* title in ``L=0`` but only for title in ``L=1`` resulting in a not translated menu item. I'm not aware of unwanted side effects but YMMV.

### DIFF
--- a/Classes/ViewHelpers/Page/Menu/AbstractMenuViewHelper.php
+++ b/Classes/ViewHelpers/Page/Menu/AbstractMenuViewHelper.php
@@ -428,9 +428,7 @@ abstract class AbstractMenuViewHelper extends AbstractTagBasedViewHelper {
 		if (0 < $getLL) {
 			$pageOverlay = $this->pageSelect->getPageOverlay($overlayPageUid, $getLL);
 			foreach ($pageOverlay as $name => $value) {
-				if (FALSE === empty($value)) {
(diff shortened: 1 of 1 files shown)


NamelessCoder:
A page named "0" or "" would not be included but I think it's enough of an edge case that we don't need to care about that.

bjo3rnf:
Thanks for merging, but why is that (before or after applying the patch):

> A page named "0" or "" would not be included but I think it's enough of an edge case that we don't need to care about that.

NamelessCoder:
Since `$var = '0';` means `empty($var) === TRUE` which means the title gets skipped (and you see the original, untranslated title instead, whatever it may be). Quite edge case ;)

bjo3rnf:
Sure, but where? This is exactly the code that has been removed...?

NamelessCoder:
Oh my... well there's a first. I read the diff the exact opposite and wrong way. I'll just go sit in this corner.

bjo3rnf:
Hehe. Things like that help me believe you _are_ a human being.

NamelessCoder:
tumblr_n6wqwgzs3f1qehu0oo1_250

PR #889, opened 2015-08-08
[BUGFIX] Page Resources need to compute sys_language_uid

### LINKED ISSUE #862: v:page.resources.fal doesn't respect language overlays
The ViewHelper ```v:page.resources.fal()```does not respect the fact that ```pages``` uses a different table in foreign languages. It should also use ```pages_language_overlay``` next to ```pages```, when the language is > 0. 

So ```{v:page.resources.fal(record: currentPage)}``` returns only images from the default language and no translated ones.

cedricziel:
Overlay table for table `pages` is in `pages_language_overlay`

Closes #862

### DIFF
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -10,6 +10,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Page\Resources;
 
 use FluidTYPO3\Vhs\ViewHelpers\Resource\Record\FalViewHelper as ResourcesFalViewHelper;
 use FluidTYPO3\Vhs\Traits\SlideViewHelperTrait;
+use TYPO3\CMS\Fluid\Core\ViewHelper\Exception;
 
 /**
  * @author Danilo Bürger <danilo.buerger@hmspl.de>, Heimspiel GmbH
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Page/Resources/FalViewHelper.php
```diff
@@ -74,11 +74,38 @@ public function render() {
 			$uid = $record['uid'];
 		}
 		if (NULL === $uid) {
-			throw new Exception('No record was found. The "record" or "uid" argument must be specified.', 1384611413);
+			throw new \Exception('No record was found. The "record" or "uid" argument must be specified.', 1384611413);
```
NamelessCoder: Please use at the very least a \RuntimeException here. Root-type \Exception is notoriously impossible to handle in tests!

Classes/ViewHelpers/Page/Resources/FalViewHelper.php
```diff
@@ -74,11 +74,38 @@ public function render() {
...
+	 * @return array
+	 */
+	public function getRecord($id) {
+		$table = $this->getTable();
+		$idField = $this->getTable() === 'pages_language_overlay' ? 'pid' : $this->idField;
+
+		$sqlIdField = $GLOBALS['TYPO3_DB']->quoteStr($idField, $table);
+		$sqlId = $GLOBALS['TYPO3_DB']->fullQuoteStr($id, $table);
+
+		return reset($GLOBALS['TYPO3_DB']->exec_SELECTgetRows('*', $table, $sqlIdField . ' = ' . $sqlId));
```
NamelessCoder: Correct me if I'm wrong, but don't we need to merge the record from `pages_language_overlay` on top of the one from `pages`? E.g. always load the default language record, then *overlay* if there is a translation...?

cedricziel: worked on my machine :)

NamelessCoder: You just need to convince me it is sufficient to switch the table name. Because I am very much not convinced that it works with any FAL TCA method and works if there is no translation of a page. I'm also not sure if even the default TCA for `pages.media` and in particular the MM would work in both 6.2 and 7.x... so there's a significant amount of doubt here!

It *looks* simple but I think it is deceiving us.

NamelessCoder:
Shouldn't this switching of tables be done only if there actually *is* a translated version of the page? Otherwise we should use the standard table?

If we don't *merge* overlay on top of the record as in https://github.com/FluidTYPO3/flux/issues/954, won't it just fail if someone tries to use this ViewHelper on a page field that *isn't* translatable?

Sorry, won't merge until those are replied to. IMHO it would cause more damage if we merge this and it begins *not showing* FAL resources that are currently shown, as opposed to incorrectly displaying the default language's resources which is sort of survivable or at least not as bad as missing assets.

NamelessCoder:
Picked with a bit of adjustment to Traits - https://github.com/FluidTYPO3/vhs/commit/59c9b3b3c9cc94e3e750d91f5f81dc3a6c9e675a

Issue #893, opened 2015-08-11
HasValidatorViewHelper doesn't support compiled rendering


thomasblass:
The HasValidatorViewHelper seems to have the same Problems under Typo 7.3 like the ExistsViewHelper in Issue #851 . 
I think there are some more ViewHelper which should be adjusted to be compatible with the Upcoming LTS Version.

xf-:
I think also conditions like isArray have problems in v7

mneuhaus:
i started working on the ConditionViewHelpers loads of them are failing in cached 7.3 currently.
will continue tonight with the rest.

xf-:
+1 :)

NamelessCoder:
Solution under review and expected to pass some time during the day tomorrow.

PR #894, opened 2015-08-13
[TASK] Make Condition ViewHelpers static compilable

mneuhaus:
many viewHelpers extending from AbstractConditionViewHelper
stopped working since 7.3 because the AbstractConditionViewHelper
is now compiled statically by default which the extending classes
now need to do too in order for the conditions to properly work
in cached instances.

There are 2 still missing/broken, because they are using the TemplateVariableViewHelperTrait
which is used quite heavily thruought vhs and thus not that easy to fix:

- IfViewHelper
- Condition\String\PregViewHelper

fixes: https://github.com/FluidTYPO3/vhs/issues/893

### DIFF
--- a/Classes/ViewHelpers/Condition/Context/IsBackendViewHelper.php
+++ b/Classes/ViewHelpers/Condition/Context/IsBackendViewHelper.php
@@ -39,21 +39,12 @@ use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractConditionViewHelper;
 class IsBackendViewHelper extends AbstractConditionViewHelper {
 
 	/**
-	 * Render method
+	 * This method decides if the condition is TRUE or FALSE. It can be overriden in extending viewhelpers to adjust functionality.
 	 *
-	 * @return string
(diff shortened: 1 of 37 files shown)


mneuhaus:
imported all fqcn i could find in the affected classes and fixed a few unit test failures.
@NamelessCoder i wasn't able to fix the IfViewHelperTest, do you have an idea what's going on there?

NamelessCoder:
@mneuhaus I get 112 test errors when testing your branch. Could you push the most recent state (force push please) so I only get the failures you're unable to fix? Thanks! :)

mneuhaus:
woot!?  when i pushed that there was 1 error :D
i'll try to take a look asap, but that might quickly stretch to the coming weekend, still in work-overload-mode

NamelessCoder:
FYI, tested it on 7.5-dev.

mneuhaus:
dropped in favor of splitted pull-requests to make it a bit more reviewable + proper tests for cached/uncached behavior.

first batch of viewhelpers waiting for review here: [TASK] Make Condition/Context ViewHelpers static compilable #906

Issue #900, opened 2015-09-07
v:page.header.link has no effect when used in templates of uncached extbase actions


JostBaron:
If the ViewHelper is used in the template of an uncached action, the `<link>` tag is added *after* the marker array has been generated (in `getPreparedMarkerArrayForPageWithUncachedObjects` in `PageRenderer.php`). Thus the link tag will not be present in the output.

It works fine in fluidpage templates and in templates of cached actions.

JostBaron:
Any workarounds on this? It's quite hard to generate the links from places other than fluid... it would at least produce a lot of doubled logic.

NamelessCoder:
There are no workarounds to my knowledge. However, I have seen a few commits about PageRenderer assets cached/uncached in the past so it could be coming. I just don't know for sure.

PR #901, opened 2015-09-07
Add info about caching to page.header.link ViewHelper

JostBaron:
Resolves issue #900


NamelessCoder:
Picked, reworded - thanks!

PR #916, opened 2015-09-24
Include encoding-option to format.tidy-Viewhelper

MoppieMop:
The default-encoding of `tidy_parse_string` is not utf8. Include option for encoding with default utf8-encoding.


NamelessCoder:
* Breaks build with invalid tidy configuration
* Commit CGL error

Sorry, must force fix this one.

PR #920, opened 2015-10-01
Check for usergroup instead for user on page-rights

MoppieMop:
Some Extensions assign only usergroup and no user - so its better to check for usergroup.

P.S. Sorry, but I don't know how to exclude my previous commit in tidy-Viewhelper. I'am git-noob.

### DIFF
--- a/Classes/Service/PageSelectService.php
+++ b/Classes/Service/PageSelectService.php
@@ -64,11 +64,14 @@ class PageSelectService implements SingletonInterface {
 	 * @return PageRepository
 	 */
 	private function createPageSelectInstance() {
-		if (FALSE === empty($GLOBALS['TSFE']->fe_user->groupData['uid'])) {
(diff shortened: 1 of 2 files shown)


MoppieMop:
Had to change the condition so "hide at login" works correctly.

bjo3rnf:
Hi @MoppieMop,

this requires some additional touch indeed ^^ but we can sort that out better on IRC I guess. You'd have to create a new branch from the current development branch first and apply your changes there. Plus the commit message should start with [BUGFIX] and describe what is actually fixed.

Cheers
Björn

MoppieMop:
Ok I'll try it. I have to create a new PR, right?

bjo3rnf:
Not necessarily, no. You can rewind your development branch, rebase, apply your changes and force push. A new branch is not mandatory (although I mentioned it above).

PR #923, opened 2015-10-07
[BUGFIX] Fix user-handling if fe_group is assigned, but no fe_user

MoppieMop:
In some special cases (with EXT:sfpipauth or other extensions) it is
possible to assign a fe_group but no fe_user. This bugfix ensures vhs
respect group assignment.

### DIFF
--- a/Classes/Service/PageSelectService.php
+++ b/Classes/Service/PageSelectService.php
@@ -72,7 +72,6 @@ class PageSelectService implements SingletonInterface {
 		if (TRUE === isset($GLOBALS['TSFE']->fe_user->groupData)) {
 			$groups = array_merge($groups, (array) array_values($GLOBALS['TSFE']->fe_user->groupData['uid']));
 		}
-		
 		$pageSelect = new PageRepository();
 		$clauses = array();
 		foreach ($groups as $group) {
--- a/Classes/Service/PageSelectService.php
+++ b/Classes/Service/PageSelectService.php
@@ -69,7 +69,7 @@ class PageSelectService implements SingletonInterface {
 		} else {
 			$groups = array(-1, 0);
 		}
-		if (FALSE === empty($GLOBALS['TSFE']->fe_user->groupData['uid'])) {
+		if (TRUE === isset($GLOBALS['TSFE']->fe_user->groupData)) {
 			$groups = array_merge($groups, (array) array_values($GLOBALS['TSFE']->fe_user->groupData['uid']));
 		}
 		
--- a/Classes/Service/PageSelectService.php
+++ b/Classes/Service/PageSelectService.php
@@ -66,13 +66,13 @@ class PageSelectService implements SingletonInterface {
 	private function createPageSelectInstance() {
 		if (TRUE === is_array($GLOBALS['TSFE']->fe_user->user)) {
 			$groups = array(-2, 0);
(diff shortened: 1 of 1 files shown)


bjo3rnf:
Well done, thanks. I can't judge the functionality though.

NamelessCoder:
I think the right way to fix this one is to change the condition that checks if a user exists.

https://github.com/FluidTYPO3/vhs/pull/923/files#diff-817458a1ca2af32b791cd0575103767cR67

If instead this checks `if (FALSE === empty($GLOBALS['TSFE']->fe_user))` or even directly `$GLOBALS['TSFE']->fe_user['groupData']`, it triggers the functionality you need, without duplicating code and without breaking anything. That would also match what then goes on inside the condition - we don't use the user record, we only use the group data.

MoppieMop:
I know what you mean, but we have to check if there is a user because of "hide at login" (group: -1) and "show at login" (group: -2). I assume that this two settings mean the user and not the (auto-assigned) group.
So my suggestion is:
```
if (TRUE === is_array($GLOBALS['TSFE']->fe_user->user)) {
	$groups = array(-2, 0);
} else {
	$groups = array(-1, 0);
}
# I assume, we do not need a check for $GLOBALS['TSFE']->fe_user->groupData['uid'] before
$groups = array_merge($groups, (array) array_values($GLOBALS['TSFE']->fe_user->groupData['uid']));

```

NamelessCoder:
The suggested fix would do the trick as far as I can tell. Before you update the pull request please verify:

* `array_values($GLOBALS['TSFE']->fe_user->groupData['uid'])` looks incorrect. Are you positive that `groupData['uid']` is an *array*? It looks like it would be an integer.
* Unless guaranteed that `$GLOBALS['TSFE']->fe_user->groupData` is *always* set, add a check around it too (validate your assumption).

If those two are in order, I'd say good to go :)

MoppieMop:
A user can have more than one group so in ```$GLOBALS['TSFE']->fe_user->groupData['uid']``` is always an array with multiple uids.
In all my tests ```$GLOBALS['TSFE']->fe_user->groupData``` was set, but I cannot guarantee it, so I added a check.

MoppieMop:
Sorry for that many commits I didn't know empty lines are not allowed.

Classes/Service/PageSelectService.php
```diff
@@ -66,10 +66,12 @@ public function initializeObject() {
 	private function createPageSelectInstance() {
 		if (TRUE === is_array($GLOBALS['TSFE']->fe_user->user)) {
 			$groups = array(-2, 0);
-			$groups = array_merge($groups, (array) array_values($GLOBALS['TSFE']->fe_user->groupData['uid']));
 		} else {
 			$groups = array(-1, 0);
 		}
+		if (TRUE === isset($GLOBALS['TSFE']->fe_user->groupData)) {
+			$groups = array_merge($groups, (array) array_values($GLOBALS['TSFE']->fe_user->groupData['uid']));
```
neufeind: nitpick: array_values always returns an array, so we might drop the additional typecast here imho. I tried array_values with diffent inputs (non-empty or empty array). In all cases it either complains there is no array or itself returns an array.

NamelessCoder:
Cherry-picked and cleaned - thanks @MoppieMop!

PR #924, opened 2015-10-08
[FEATURE] runtime caching of unserialized array, possibility to retur…

jgreiner:
…n array

- saving unserialized array in static var to avoid multiple unserialisations in runtime
- possibility to return array (dots removed)
- trimming string result
- minimal call <v:variable.extensionConfiguration /> returns whole extension configuration as array

### DIFF
--- a/Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
+++ b/Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
@@ -23,6 +23,7 @@ use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper;
  * Returns setting 'foo' from extension 'bar' located in ext_conf_template.txt
  *
  * @author Harry Glatz <glatz@analog.de>
+ * @author Jochen Greiner <greiner@analog.de>
  * @author Cedric Ziel <cedric@cedric-ziel.com>
  * @package Vhs
  * @subpackage ViewHelpers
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
```diff
@@ -30,25 +31,39 @@
...
 			$extensionName = $this->controllerContext->getRequest()->getControllerExtensionName();
 			$extensionKey = GeneralUtility::camelCaseToLowerCaseUnderscored($extensionName);
 		}
-
 		if (FALSE === array_key_exists($extensionKey, $GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'])) {
 			return NULL;
 		}
 
-		$extConf = unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'][$extensionKey]);
+		if (empty(self::$EXT_CONF[$extensionKey])) {
```
neufeind: I'd use isset or array_key_exists instead here

Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
```diff
@@ -30,25 +31,39 @@
...
+			self::$EXT_CONF[$extensionKey] = GeneralUtility::removeDotsFromTS(unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'][$extensionKey]));
+		}
+		if (!$name) {
```
neufeind: if (NULL === name) imho.

neufeind:
The optimization would be fine with me. But imho we can't really remove the dots or trim a string (BC-break).

NamelessCoder:
@neufeind This change (minus the dot stripping) looks good to me. Feel free to cherry-pick, amend (and correct commit subject) then push directly!

cedricziel:
Manually resolved the config and implemented a simple cache with b3d089848c598845b0f2c923c5211e482c740e60

Please review again to make sure I didn't do bad things while playing the keyboard monkey.

Oh and thank you @jgreiner for the idea and the first patch! :)

cedricziel:
Ahh. Dang. I was so into PSR2 already, I didn't remember the huge difference. Opened #978 to apply the new psr2 esque code style to the whole library.

NamelessCoder:
Little too quick on this one. Subject is bad and it's based on a commit quite a while back.

And I can't really say I agree with fixing this code style issue by converting the rest of the repository to the incorrect code style ;)

cedricziel:
Hehe. Agreed. I'll leave the other one open, but revert the merge.

NamelessCoder:
For the reworked version, please take care of previous code comments plus an additional one from me: since abbreviations are discouraged, `$res` should be renamed to something like `$result`.

PR #944, opened 2015-11-07
[TASK] Use first srcset width as default fallback src

### LINKED ISSUE #937: Responsive image ViewHelpers ahving wrong fallback src
When using the Media.image ViewHelper we discovered that the smallest srcset option is uses as a fallback for the src attribute.
In internet Explorer, where srcsets are not used, we get the smallest image for the Fallback. This is not desired in most cases since IE is mostly used on desktops.

I propose to remove the [ksort in the trait](https://github.com/FluidTYPO3/vhs/blob/development/Classes/Traits/SourceSetViewHelperTrait.php#L50), so we use the first defined srcset. In this case we are able to select the src by just order the srcsets accordingly.

mneuhaus:
fixes #937

### DIFF
--- a/Classes/Traits/SourceSetViewHelperTrait.php
+++ b/Classes/Traits/SourceSetViewHelperTrait.php
@@ -47,8 +47,8 @@ trait SourceSetViewHelperTrait {
 		}
 		$tag->addAttribute('srcset', implode(',', $srcsetVariants));
 
+		$tag->addAttribute('src', reset($imageSources));
 		ksort($imageSources);
-		$tag->addAttribute('src', current($imageSources));
 
 		if ('BE' === TYPO3_MODE) {
 			FrontendSimulationUtility::resetFrontendEnvironment();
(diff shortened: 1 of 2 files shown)


Classes/Traits/SourceSetViewHelperTrait.php
```diff
@@ -47,8 +47,8 @@ public function addSourceSet($tag, $src) {
 		}
 		$tag->addAttribute('srcset', implode(',', $srcsetVariants));
 
-		ksort($imageSources);
 		$tag->addAttribute('src', current($imageSources));
```
neufeind: Should we maybe use reset() (returns the first element then) instead of current()? Just "to be sure" which element is there current one here.

mneuhaus: yes, ... of course ... that's what i planned to do actually before composer/hotel wifi codeblocked be for ~1.5h ...

PR #945, opened 2015-11-08
[FEATURE] ExtensionConfiguration-VH: Allow fetching subpart of config

neufeind:
The parameter "name" existed before and was not yet optional.
It is now replaced by an optional parameter "path" to specify a path to
fetch a subpart of configuration from.

### DIFF
--- a/Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
+++ b/Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
@@ -10,6 +10,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Variable;
 
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3\CMS\Fluid\Core\ViewHelper\Exception;
 
 /**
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Variable/ExtensionConfigurationViewHelper.php
```diff
@@ -18,24 +18,46 @@
...
 	/**
-	 * @param string $name
+	 * @param array $source TypoScript-array with dots: $source['foo.']['bar.']['baz']
+	 * @param string $path
+	 * @return mixed
+	 */
+	protected function extractFromArrayByPath($source, $path) {
+		$result = $source;
+		$pathParts = explode('.', $path);
+		$pathParts = array_diff($pathParts, ['']);
```
NamelessCoder: Not yet converted to `[]` syntax, please use `array()` until we make the full switch. And add space after `foreach` before `(` - everything else looks spot on and should support the fairly odd case of having to access these special dot-suffixed arrays using a proper Fluid syntax.

NamelessCoder:
Since this involves an API change that breaks - should we instead have both "path" and "name" arguments and make both optional, then throw a ViewHelperException if neither is specified and warn about deprecation via logging if "name" is used? I don't even know how many people use this ViewHelper but there's at least one live site that depends on it (original VH author's project).

PR #951, opened 2015-11-12
[DOC] Update changelog for upcoming version

InvisibleKind:


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 upcoming
 ------------------
 
+- [TYPO3 7 LTS supported](https://github.com/FluidTYPO3/vhs/commit/ab893323e2cac57ca32b22efe20a7ad67c3e7bff)
+
 - :exclamation: No more testing for PHP 5.4
 	- VHS still supports TYPO3 6.2, which can be run on PHP 5.4, but it is advised to upgrade your PHP to 5.5 (which is also supported by 6.2) at least
```
NamelessCoder: I think this may be incorrect now. IIRC 7.4 is minimum supported version now and we test on 7.6, not 6.2. The PHP info is still correct.

InvisibleKind: Okay, I'll change that. Just checked the info from ext_emconf.php, where 6.2 is still the lowest supported version.

PR #954, opened 2015-11-20
[TASK] Ignore some files/directories on archive export

mbrodala:
In case the extension is pulled via Composer, a few files and directories can
be excluded to reduce the amount of transferred data and disk space usage.


mbrodala:
Not sure about `Resources/Private/Schemas/Vhs.xsd` so I kept it for now.

cedricziel:
Thx!

IMHO leaving schema in there is fine.

mbrodala:
Thanks. :-)

PR #956, opened 2015-11-25
[DOC] Image example for vhs:resource.record.fal

cweiske:


cedricziel:
Not entirely. Time will tell :)

Thx a bunch!

Issue #927, opened 2015-10-13
Async/defer for asset.script


InvisibleKind:
Small feature request to add these two attributes to v:asset.script.

bjo3rnf:
See vhs#625

bjo3rnf:
But: As far as I can tell this would require to exchange tag builder with a lightweight custom implementation to render the attribute without a value (HTML5 style) and force attribute ``standalone`` to be set on those assets with either ``async`` or ``defer``.  No big deal. Any obligations?

bjo3rnf:
Closing this one after discussing the implications that are still existing and impossible to workaround in the current implementation.

NamelessCoder:
It's sad to say but I completely agree - although we don't like doing it, we have to give up in advance on this case. There's just too much standing in the way of it working.

Issue #807, opened 2015-04-07
[RFC][FEATURE] Wraps for Assets


cedricziel:
I think we need wraps for Assets. - The one or the other way. TS Conditions suck for conditional stylesheets for example.

Because the rewrite of the asset part didnt happen at all (my fault _ducksandruns_), we need to take this a bit further.

In my opinion we need wraps for:

* [ ] groups
* [ ] header parts
* [ ] footer parts

Can you please suggest places where you think this would make sense and share general thoughts? Technical details apart please :)

ping: @outdoorsman @NamelessCoder @bjo3rnf

NamelessCoder:
I'm all for supporting a wrapping of sorts, but I absolutely think we should be separating this to a new extension first. And if I am not mistaken, such features would be supported by for example Assetic?

I would do this as follows:

* Create the new extension and adopt the ViewHelpers and their parameter namings, but have them create Assetic objects/config instead.
* Create in this new extension a temporary API that will consume VHS asset objects.

This will allow us to properly deprecate the asset feature of VHS giving it however long is required. The benefit here is that any features we create _after_ this point will be created in the proper way in the proper extension - if we don't, we simply add to the pile of features we have to replicate in or delegate to whatever asset handling dependency we introduce.

And we'd like to do less, not more. IMHO the ideal way to achieve that is to delegate to a third-party dependency (one that's considered a standard). This extension becomes a wrapper for the third-party dependency and VHS becomes a wrapper for this extension; either through an actual dependency or a silent one with some friendly messaging, because the VHS wrapper technically becomes deprecated immediately.

BenjaminBeck:
+1 for using Assetic
And there is one thing that recently happened to me:
I added some css with v:asset.style inside a content element that is not present on every page.
The behaviour here was, that this style was merged with the main stylesheet and created a completely new file (which broke caching..). Maybe this -default behaviour- can be improved?

cedricziel:
I would like to give up on this, as I consider a clean asset pipe decoupled from the cms state of the art.

Let me know what you think.

bjo3rnf:
https://youtu.be/OTLoEr21raA

NamelessCoder:
Drop it like it's hot ;)

cedricziel:
Rocked to the bottom :)

PR #957, opened 2015-11-27
[ADD] Adding media Parameter to Assets for Stylesheet

BenjaminOberhauser:


NamelessCoder:
Hi Benjamin,

This will unfortunately cause problems. We've had a few discussions over the time about media and related features and the core of the issue is: your assets get merged and that means each merged chunk must always use the same media and other properties. If they don't, merging happens incorrectly. Because of this, assets have to be grouped once more based on an arbitrary number of different properties - and we've opted to not open that can of worms.

See:

* https://github.com/FluidTYPO3/vhs/issues/927
* https://github.com/FluidTYPO3/vhs/issues/807

So I'm very sorry but we have to reject this one. We can't begin implementing these without huge changes in the asset handling logic - and we would much rather spend that time on (just for example) implementing Assetic with a wrapper to let it "import" current VHS assets. Since composer support is now pretty good I expect this will happen at some point, perhaps even natively supported by TYPO3. Until that time you can either use TypoScript to generate the style tag manually (not using VHS), or you can use `v:asset` with the entire `<style>` tag manually inserted as tag content.

PR #962, opened 2015-12-14
[TASK] Cleanups in FalViewHelper

smichaelsen:
Fix phpDoc annotations, yoda-style comparisons and other small
improvements.

### DIFF
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -9,6 +9,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Resource\Record;
  */
 
 use FluidTYPO3\Vhs\Utility\ResourceUtility;
+use TYPO3\CMS\Core\Resource\FileReference;
 use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Database\DatabaseConnection;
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -123,19 +120,19 @@ public function getResources($record) {
...
-			if (FALSE === empty($referenceUids)) {
-				foreach ($referenceUids as $referenceUid) {
-					try {
-						// Just passing the reference uid, the factory is doing workspace
-						// overlays automatically depending on the current environment
-						$images[] = $this->resourceFactory->getFileReferenceObject($referenceUid);
-					} catch (ResourceDoesNotExistException $exception) {
-						// No handling, just omit the invalid reference uid
-						continue;
+				$images = array();
```
NamelessCoder: Moving declaration of $images in here may cause the variable to be undefined (or at least not an array) which would cause the final loop to break with a fatal PHP error. It should be okay to move the declaration outside of the condition so it always gets initialised but only filled if referenceUids is not empty.

Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -74,11 +75,11 @@ public function getResource($fileReference) {
...
 	/**
-	 * Fetch a fileRefernce from the file repository
+	 * Fetch a fileReference from the file repository
 	 *
-	 * @param $table name of the table to get the file reference for
-	 * @param $field name of the field referencing a file
-	 * @param $uid uid of the related record
+	 * @param string $table name of the table to get the file reference for
+	 * @param string $field name of the field referencing a file
+	 * @param int $uid uid of the related record
```
NamelessCoder: Please use `integer` here - everything else looks good!

PR #974, opened 2016-01-13
[TASK] Remove deprecated argument

### LINKED ISSUE #964: [BUG] Asset/AbstractAssetViewHelper
Hello

I think there is a possible bug when property `movable` is used by `<v:asset.script />` and `<v:asset.style />`. Condition from line 361 always returns `TRUE` and it overwrites this setting to `TRUE` (default value for `allowMoveToFooter`)

Tested with vhs 2.4.0

bjo3rnf:
Fixes #964


cedricziel:
Thank you! :)

Issue #978, opened 2016-01-20
[TASK] Reformat to PSR-2 alike core standard


cedricziel:

bjo3rnf:
Done in 95e2547f.

PR #979, opened 2016-01-20
Apply PSR-2 Code style

cedricziel:
Since core uses it, there's no reason why we shouldn't adopt it.

Closes: #978


cedricziel:
ping @NamelessCoder :

Any objections if I'd re-apply PSR-2 with the core standard to all of our code base this weekend? I think it's best to just adopt the core style and its config so we don't have to take care about those issues anymore.

NamelessCoder:
No objections from me - as long as you prepare them as PRs so we can merge them all at the same time. You may want to wait until at least tomorrow though, I might make a few releases later tonight.

cedricziel:
I'm super fine with this. I'll do so tomorrow then.

cedricziel:
Alright - I think this one's good to go, once FluidTYPO3/fluidtypo3-development#2 is in.

bjo3rnf:
Let's do this.

cedricziel:
Sure as hell. Please check https://github.com/FluidTYPO3/fluidtypo3-development/pull/2

NamelessCoder:
Done, but with raw PSR-2 standard instead of TYPO3CMS one.

PR #980, opened 2016-01-21
[TASK] Add enableJsApi parameter to YoutubeViewHelper

wtfred:
So here is the PR with one commit ^^


cedricziel:
Thx for taking the time!

wtfred:
No problem, thank you for all the rest of the extension ^^

cedricziel:
Very apprecated, thx a bunch :)

PR #981, opened 2016-02-02
[BUGFIX] Fix the cast of the playerId argument

wtfred:
Hi,

An ultra small PR just to remove the casting of the argument "playerId" as an integer.
On this page https://developer.vimeo.com/player/js-api, we can see that "player_id" should be a string :)


cedricziel:
Alright! Tank you for your contribution :)

wtfred:
You're welcome !

Issue #865, opened 2015-07-22
v:resource.image does not respect fully qualified uri's


mischka:
I use http://typo3.org/extensions/repository/view/aus_driver_local_cdn as File Storege Driver.
So the image uri is http://cdn.mydomain.tld/my-image.jpg but the `ImageViewHelper` encodes the uri so my image.source is `http%3A//cdn.mydomain.tld/my-image.jpg` . The problem is in the `preprocessImages` and the `preprocessSourceUri`-method in the `FluidTYPO3\Vhs\ViewHelpers\Resource\AbstractImageViewHelper`, there is no check for a fully qualified uri

    $imageSource = $GLOBALS['TSFE']->absRefPrefix . GeneralUtility::rawUrlEncodeFP($imageInfo[3]);

to

    if(GeneralUtility::isValidUrl($imageInfo[3])) {
        $imageSource = $imageInfo[3];
    } else {
        $imageSource = $GLOBALS['TSFE']->absRefPrefix . GeneralUtility::rawUrlEncodeFP($imageInfo[3]);
    }

PR #991, opened 2016-02-19
[BUGFIX] SourceViewHelper should respect fully qualified uris

kersten:
Same as #865


bjo3rnf:
Manually picked, CGL issues fixed and merged. Thanks for contributing.

PR #995, opened 2016-03-02
[BUGFIX] Fixed addQueryString behaviour in LanguageMenuViewHelper

andypa:
https://docs.typo3.org/typo3cms/TyposcriptReference/Functions/Typolink/Index.html#typolink <-- addQueryString is a boolean with the property method which needs to be set.


NamelessCoder:
Thanks Andy!

PR #1000, opened 2016-03-17
[TASK] Remove deprecated dependency

bjo3rnf:
I just noticed that vhs causes an entry in deprecation.log of TYPO3 7.6.4


cedricziel:
How do you feel, Mr 1000?

bjo3rnf:
> How do you feel, Mr 1000?

I'm flabbergasted and proud to have provided such an important commit codewise.

PR #1003, opened 2016-03-24
[TASK] Remove v:switch and v:case

NamelessCoder:
Can be safely migrated by string-replacing `v:switch` to `f:switch` and `v:case` to `f:case`.


cedricziel:
That woul mean vhs 3.0 - is there anything we can release upfront?

NamelessCoder:
That is exactly what that would mean :)

We can make a bugfix release for 2.4.1 if necessary.

NamelessCoder:
We could also make class aliases. I'll do that!

cedricziel:
Alright - great idea!

cedricziel:
Are they 100% compatible argument-wise?

NamelessCoder:
> Are they 100% compatible argument-wise?

Hmm, now that you mention it... they're not. We do have to remove these without aliases, and the migration description needs to include argument differences.

cedricziel:
Agreed. :)

TheNaderio:
Someone shold mark the v:switch as DEPRECATED in the viewhelper reference at:
https://fluidtypo3.org/viewhelpers/vhs/master/SwitchViewHelper.html

PR #1007, opened 2016-03-31
[BUGFIX] make sure we use pages_language_overlay table instead of pages table if the given record is a overlay record fixes #996

### LINKED ISSUE #996: using v:page.resources.fal on translated pages
Hi there,
I'm using vhs 2.4.0 with TYPO3 7.6.4.

I use v:page.resources.fal with a page uid to get the Image related to this page via FAL in the Page-Properties. So far so good.    

When I use it with a translated version of a page - but same pageUid - I get a wrong Image.  
I get the Fal-Image of that Page that has the uid of the translation (pages_language_overlay) of the actual page. 

Is this somehow understandabele? 

kind regards
Thomas

torvitas:
make sure we use pages_language_overlay table instead of pages table if the given record is a overlay record

### DIFF
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -108,9 +108,7 @@ class FalViewHelper extends ResourcesFalViewHelper {
 	 */
 	public function render() {
 		$record = $this->arguments['record'];
-        if (isset($record[self::defaultOverlayField]) &&
-            $record[self::defaultOverlayField]
-        ) {
+        if (!empty($record[self::defaultOverlayField])) {
             $this->arguments['table'] = self::defaultOverlayTable;
             $this->arguments['uid'] = $record[self::defaultOverlayUidField];
         }
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -24,6 +24,9 @@ class FalViewHelper extends ResourcesFalViewHelper {
 
 	const defaultTable = 'pages';
 	const defaultField = 'media';
+    const defaultOverlayField = '_PAGES_OVERLAY';
+    const defaultOverlayUidField = '_PAGES_OVERLAY_UID';
+    const defaultOverlayTable = 'pages_language_overlay';
 
 	/**
 	 * @var string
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Page/Resources/FalViewHelper.php
```diff
@@ -105,6 +108,12 @@ public function getActiveRecord() {
 	 */
 	public function render() {
 		$record = $this->arguments['record'];
+        if (isset($record[self::defaultOverlayField]) &&
+            $record[self::defaultOverlayField]
```
bjo3rnf: Please put it in a single line and use ``!empty($record[self::defaultOverlayField])``

bjo3rnf:
Hi Sascha,

thanks for your contribution. Please see my comment and apply a more verbose commit message (flipping comment and commit message would be perfect).

Cheers
Björn

cedricziel:
Hi @torvitas ,

Do you plan on improving on the issues?

Cheers,
Cedric

NamelessCoder:
Looking at this again I think the appropriate solution is to apply the language overlay in `getRecord` so that method returns the translated/overlayed record, and as previously mentioned in `getSideRecords` as well. It may also require that we change the methods that resolve related records, to use the original UID column instead of `uid` when the intention is to use the parent/untranslated record's UID.

Sengchheang:
I cannot get luck from the current patch because the code block below defeats the purpose of the fix.
Oh reversely, the patch supports only once case but no support when fallback 

```
if (NULL === $uid) {
			if (NULL === $record) {
				$record = $this->getActiveRecord();
			}
			$uid = $record['did'];
		}
```

Here is my setup 
`<v:page.resources.fal as="imagesSlider">`

So to get this fixed, I need to re-arrange quite a bit of ordering.
Here is my whole block

```
public function render() {
		$uid = $this->arguments['did'];

		if (NULL === $uid) {
			$record = $this->arguments['record'];

			if (NULL === $record) {
				$record = $this->getActiveRecord();
			}

			$uid = $record['did'];

			if (!empty($record[self::defaultOverlayField])) {
				$this->arguments['table'] = self::defaultOverlayTable;
				$this->arguments['uid'] = $record[self::defaultOverlayUidField];
				$uid = $record[self::defaultOverlayUidField];
			}
		}

		if (NULL === $uid) {
			throw new Exception('No record was found. The "record" or "uid" argument must be specified.', 1384611413);
		}
		$resources = $this->getSlideRecords($uid);

		return $this->renderChildrenWithVariableOrReturnInput($resources);
	}
```

Here below is the original

```
public function render() {
		$record = $this->arguments['record'];
        if (!empty($record[self::defaultOverlayField])) {
            $this->arguments['table'] = self::defaultOverlayTable;
            $this->arguments['uid'] = $record[self::defaultOverlayUidField];
        }
		$uid = $this->arguments['did'];
		if (NULL === $uid) {
			if (NULL === $record) {
				$record = $this->getActiveRecord();
			}
			$uid = $record['did'];
		}
		if (NULL === $uid) {
			throw new Exception('No record was found. The "record" or "uid" argument must be specified.', 1384611413);
		}
		$resources = $this->getSlideRecords($uid);
		return $this->renderChildrenWithVariableOrReturnInput($resources);
	}
```

PR #1008, opened 2016-03-31
fixed query string added get method

lukasgurschler:


bjo3rnf:
Fixed in ad215751. Thanks nevertheless!

Issue #1009, opened 2016-04-01
condition.string.preg argument "as" missing


hacksch:
I miss the argument "as" in the viewhelper declaration.
In line 101 the argument will be used. But its not possible to set them without declaration in method initializeArguments.

bjo3rnf:
Hi Martin,

this is actually some special case. ``as`` should be registered with TemplateVariableViewHelperTrait::registerAsArgument() which is missing here but unfortunately the use case of the as argument is different to the other viewhelpers as it should contain the matches instead of the viewhelper content. Thus the trait can't be implemented like this. I'll have a look at it.

Cheers
Björn

bjo3rnf:
To quote @NamelessCoder:

> "fixed" by refactoring the preg-related ViewHelpers to two proper scopes - see #1023. Thanks for bringing this to our attention!

PR #1010, opened 2016-04-01
#1009 condition.string.preg argument "as" missing

hacksch:


xf-:
hi,
it should be already registered via TemplateVariableViewHelperTrait
https://github.com/FluidTYPO3/vhs/blob/development/Classes/Traits/TemplateVariableViewHelperTrait.php

hacksch:
Maybe my implementation is not correct...
The use case is to use the vhs view helpers in another extension. Included by {namespacev =...}.
It looks like the TemplateVariableViewHelperTrait wasn't called in that case to register the argument because i get an exception.
What can i do?

NamelessCoder:
"fixed" by refactoring the preg-related ViewHelpers to two proper scopes - see #1023. Thanks for bringing this to our attention!

PR #1012, opened 2016-04-11
[BUGFIX] Register as argument and assign matches

### LINKED ISSUE #1009: condition.string.preg argument "as" missing

bjo3rnf:
This patch fixes #1009 in a way that it registers the original as argument which was missing after introducing the TemplateViewHelperTrait and enables the usage of additional template variables besides ``$as``. The latter is open for discussion bc I am not sure if it is acceptable thus the two commits.


bjo3rnf:
Better solution incoming.

PR #1013, opened 2016-04-11
[BUGFIX] SourceSet to work with treatIdAsReference

### LINKED ISSUE #1011: [BUG] SourceSetViewHelperTrait does not work with Media/ImageViewHelper and treatIdAsReference=true
The Problem is ``vhs/Classes/Traits/SourceSetViewHelperTrait.php`` ``~L:44``
where ``$this->getImgResource`` does use the ``$src`` parameter which comes from
``vhs/Classes/ViewHelpers/Media/ImageViewHelper.php``  ``~L:77`` (``$this->mediaSource``) and is the path to the already processed image and not as required the file-reference-uid.

BenjaminBeck:
This fixed #1011

### DIFF
--- a/Classes/Traits/SourceSetViewHelperTrait.php
+++ b/Classes/Traits/SourceSetViewHelperTrait.php
@@ -35,6 +35,9 @@ trait SourceSetViewHelperTrait {
 		$format = $this->arguments['format'];
 		$quality = $this->arguments['quality'];
 		$treatIdAsReference = (boolean) $this->arguments['treatIdAsReference'];
+		if (true === $treatIdAsReference) {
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hi Benjamin - looks okay (I'm not 100% comfortable with the internals of FAL/image viewhelpers but from reading, it would certainly "tread reference as ID") but there's a CGL issue to take care of:

https://travis-ci.org/FluidTYPO3/vhs/jobs/122185906#L322

After that I think we can merge this.

bjo3rnf:
Manually picked and merged. Thanks Benjamin.

PR #1023, opened 2016-04-16
[TASK] Refactor regular expression related viewhelpers

bjo3rnf:
This patch removes ``v:condition.string.preg`` and ``v:format.regularExpression`` in favor of ``v:format.pregReplace`` and ``v:variable.pregMatch``.


NamelessCoder:
Classes look good - unit tests need to be renamed/moved/deleted.

PR #1027, opened 2016-04-27
[TASK] Reduce number of db queries

bjo3rnf:
This patch slightly reduces the number of db queries fx when generating menus by enabling ``PageService::hidePageForLanguageUid`` to accept full page records.

### DIFF
--- a/Classes/Service/PageService.php
+++ b/Classes/Service/PageService.php
@@ -140,20 +140,23 @@ class PageService implements SingletonInterface {
 	}
 
 	/**
-	 * @param integer $pageUid
+	 * @param array|integer $page
 	 * @param integer $languageUid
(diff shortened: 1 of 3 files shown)


Classes/Service/PageService.php
```diff
@@ -140,20 +140,23 @@ protected function getPageConstraints(array $excludePages = array(), $includeNot
 	}
 
 	/**
-	 * @param integer $pageUid
+	 * @param mixed $page
```
cedricziel: no mixed please :) `int|array` it is

bjo3rnf: and by 'int' you mean 'integer', right? ^^ So be it.

cedricziel:
Cheers!ölf

PR #1046, opened 2016-06-02
[FEATURE] Add new MimetypeIconViewHelper to return the icon font CSS …

danielkestler:
…class for a given file extension.

Note: this is just a proposal, please let me know if you think it is not generic enough etc.

### DIFF
--- a/Classes/ViewHelpers/Media/MimetypeIconViewHelper.php
+++ b/Classes/ViewHelpers/Media/MimetypeIconViewHelper.php
@@ -1,0 +1,91 @@
+<?php
+namespace FluidTYPO3\Vhs\ViewHelpers\Media;
+
+/*
+ * This file is part of the FluidTYPO3/Vhs project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+/**
+ * Returns an icon font CSS class for a given file extension.
+ * The default mimetype map provides the class suffix for FontAwesome.
+ *
+ * ### Examples
+ *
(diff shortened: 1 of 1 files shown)


xf-:
i think providing static list of available on 1 of 100 icon-fonts (and customization tools) and you only provide a small selection of icons and mapping. Additionally this needs maintenance for new icons.

danielkestler:
Yeah, you're right. That's what the map-parameter is for ... but I think it might indeed need a bit more abstraction or some more supported icon fonts? But FontAwesome is really one of the most popular ones.

NamelessCoder:
We would definitely need a map or similar argument for this one to work, generally speaking we should avoid dependency-specific ViewHelpers. There *is* an exception currently for which I'm to blame - `v:format.markdown` (for which you can only switch the markdown implementation by changing the mapped `markdown` command in the TYPO3 install tool).

That being said I'm not sure this one would have a very long life. Basically it is a VH to get an index of an array by using another variable, which in future Fluid versions is pretty easy to solve:

```xml
<!-- $mimes = ['foo' => 'bar-icon', 'baz' => 'baz-icon'] -->
<span class="icon fa fa-{mimes.{file.mimetype}}"></span>
```

It's such a general use case that I personally would opt to teach and advertise that sub-variable addressing approach in favor of dedicated ViewHelpers that do similar mapping.

NamelessCoder:
Sorry to reject this one, but standard arrays to define the exact mime types you need, combined with dynamic access based on mime type (or whichever other type you wish), is the solution which does not restrict itself to a particular collection of icons or set of mime types - and does not require that we maintain this association when new icon sets or mime types are added. Or when existing ones are renamed which would start the downward spiral of adding versioned mapping arrays and version arguments.

However: this type of ViewHelper might make a lot of sense if it is included within a library package that adds FontAwesome assets for a site. That would allow the ViewHelper to be perfectly in sync with the FontAwesome version. We don't maintain or plan to create such a package but if you know of one, you should definitely suggest this to them more or less as-is (perhaps I would switch the extension-to-icon mapping to a full mimetype-to-icon map first).

PR #1054, opened 2016-06-15
[TASK] Deprecate typolink viewhelpers, closes #1051

### LINKED ISSUE #1051: [TASK] Remove v:uri.typolink
and v:link.typolink.
In TYPO3 7.6 there is a viewhelper f:link.typolink

danielkestler:
I have marked the viewhelpers deprecated and offloaded everything to the Fluid core viewhelpers.


Classes/ViewHelpers/Uri/TypolinkViewHelper.php
```diff
@@ -21,23 +23,27 @@
  * For more info on the typolink function, please consult the offical core-documentation:
  * http://docs.typo3.org/typo3cms/TyposcriptIn45MinutesTutorial/TypoScriptFunctions/Typolink/Index.html
  *
+ * DEPRECATED: Use TYPO3\CMS\Fluid\ViewHelpers\Uri\TypolinkViewHelper instead
+ *
  * ### Examples
  *
  *     <!-- tag -->
  *     <v:uri.typolink configuration="{typoLinkConfiguration}" />
- *     <v:uri.typolink configuration="{object}">My LinkText</v:uri.typolink>
```
bjo3rnf: Does this mean providing the linktext like so is no longer supported? This would create a BC break if so.

cedricziel: That would indeed be a BC break. If you really want to pull in a transition layer, please support all the previous arguments.

danielkestler: This was meant only as a correction to the documentation – in the v:uri.typolink viewhelper, adding a linktext or a class does not make sense at all, since it's only returning the url?

bjo3rnf: Oh! Yes, that's absolutely correct. Sorry.

cedricziel: Dangit. You're right. My apologies for this one ^^

danielkestler: Don't mind ;)

NamelessCoder:
Good catch btw., re: renderContent on the URI-version!

PR #1055, opened 2016-06-16
[BUGFIX] '(boolean) $this->getCurrentLanguageUid() === 0' is always FALSE in Page/Resources/FalViewHelper

froehlit:
The result of the comparison is not castetd, instead the result of ```isDefaultLanguage()`` is casted to boolean. It always runs on a comparison between bool and int.


bjo3rnf:
Manually picked and merged. Thanks for spotting this! I took the freedom to remove the type cast completely as it doesn't make sense anyway.

PR #1057, opened 2016-06-17
[BUGFIX] Collection viewhelper always returning NULL

squidfunk:
Fixes the problem, that the resource collection viewhelper returns the return value of `loadContents`, which is `void`, so the result is always `NULL`.

```
{v:resource.collection(uid: uid) ->  v:variable.set(name: 'records')}
```

### DIFF
--- a/Classes/ViewHelpers/Resource/CollectionViewHelper.php
+++ b/Classes/ViewHelpers/Resource/CollectionViewHelper.php
@@ -49,7 +49,7 @@ class CollectionViewHelper extends AbstractResourceViewHelper {
 			/** @var \TYPO3\CMS\Core\Collection\AbstractRecordCollection $collection */
 			$collection = $this->collectionRepository->findByUid($uid);
 			if (NULL !== $collection) {
-				return $collection->loadContents();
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Resource/CollectionViewHelper.php
```diff
@@ -50,9 +50,8 @@ public function render($uid) {
 			$collection = $this->collectionRepository->findByUid($uid);
 			if (NULL !== $collection) {
 				return $collection->loadContents();
-			} else {
-				return NULL;
 			}
+			return $collection;
```
bjo3rnf: Please correct me if I'm wrong but this would still return nothing when $collection is not null, right?

cedricziel: Just remove the return from line 52, let it load the collection and return whatever is in `$collection`

squidfunk:
Sorry, of course you're right. I fixed it in my local copy, but got it wrong in my pull request - should be fixed now.

bjo3rnf:
Thanks Martin!

Issue #896, opened 2015-08-21
v:resource.record.fal does not respect workspaces at all


mkrappitz:
Given the following EXT:fluidcontent content element for example:

```html
{namespace flux=FluidTYPO3\Flux\ViewHelpers}
{namespace v=FluidTYPO3\Vhs\ViewHelpers}

<f:layout name="Content" />

<f:section name="Configuration">
	<flux:form id="image" label="Image" description="Image" options="{Fluidcontent: {sorting: 05}, group: 'FCE'}">
		<flux:field.inline.fal name="images" label="Image" minItems="1" maxItems="1" collapseAll="TRUE" />
	</flux:form>
</f:section>

<f:section name="Preview">
	...
</f:section>

<f:section name="Main">
	<f:for each="{v:content.resources.fal(field: 'images')}" as="image">
		<f:image treatIdAsReference="TRUE" src="{image.id}" alt="{image.alternative}" title="{image.title}" />
	</f:for>
</f:section>
```

it all works fine in the live workspace editing and creating such elements. Once in a draft workspace editing existing elements or creating new ones of this kind, v:content.resources.fal which internally triggers v:resource.record.fal fails to handle the sys_file_reference records correctly. It seems very obvious to me why that is the case once you look into the source of EXT:vhs/Classes/ViewHelpers/Resource/Record/FalViewHelper.php method getResources($record). There is no handling of workspaces in the sql query on line 57 at all, which results in the call of $this->getResource($row['uid']) on line 61 to fail and the $resources result becoming invalid.

Tested on TYPO3 CMS 6.2.14 and 7.4.0
EXT:vhs 2.3.3
EXT:fluidcontent 4.3.1

mkrappitz:
It might also be the case that this error would not be existing if flux #898 would get fixed. But this is just speculation. Anyway I think these issues are connected.

tiwden:
Not sure weather this is still helpfull, but we expirienced some problematic behaviour if the viewhelper is used in backend scope, for example in preview section. If there is a Live Record set to VersionState::DELETE_PLACEHOLDER the Record is still fetched and shown in the preview section.

to Prevent this, we changed 

`
foreach ($images as $file) {
  $resources[] = $this->getResource($file);
}`

to

`
foreach ($images as $file) {
  if($file->getProperty('t3ver_state') !== VersionState::DELETE_PLACEHOLDER){
    $resources[] = $this->getResource($file);
  };
}`

[...]

PR #1059, opened 2016-06-23
[BUGFIX] Localise every getRecord call in v:page.resources.fal

NamelessCoder:
Apply localisation so that inherited render() method will retrieve a localised version of the record. Also removes the overridden render() method on v:page.resources.fal since this method was the same as the inherited one. Now, the inherited method calls getRecord which then applies the localisation.

Close: #1007
Close: #996


NamelessCoder:
Fix for #896 follows but depends on this change.

PR #1060, opened 2016-06-23
[BUGFIX] Adapt to recent changes in image viewhelpers

### LINKED ISSUE #933: PdfThumbnailViewHelper renders no thumbnails
In the PdfThumbnailViewHelper-Class the render-method called the parent::render-Method with some parameters.
But ImageViewHelper->render has no parameters........
So the new pdf-thumbnail src will not transfered to ImageViewHelper.

bjo3rnf:
This is not much more of a hotfix to make PDF thumbnail generation work again. Some arguments are deprecated due to changes in image viewhelpers. Closes #933.

### DIFF
--- a/Classes/ViewHelpers/Media/Image/AbstractImageViewHelper.php
+++ b/Classes/ViewHelpers/Media/Image/AbstractImageViewHelper.php
@@ -109,11 +109,12 @@ abstract class AbstractImageViewHelper extends AbstractMediaViewHelper
     }
 
     /**
+     * @param string|null $imageSource
      * @throws Exception
      */
-    public function preprocessImage()
+    public function preprocessImage($imageSource = null)
     {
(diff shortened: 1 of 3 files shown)


NamelessCoder:
One minor wish: could you put the "imageSource" argument in initializeArguments (pragma will be to avoid render method arguments) - and refactor the image generating method away from "render" on parent to a dedicated method if that's necessary.

bjo3rnf:
> and refactor the image generating method away from "render" on parent to a dedicated method if that's necessary

I think this the better solution as the 'imageSource' argument is not really a viewhelper argument. I'll take care of that shortly.

NamelessCoder:
Seems there's a single GCL issue - line length are now hard restricted to 120 chars. Other than that, looks good to me!

Issue #1062, opened 2016-06-25
v:page.menu login restricted access


7elix:
Hi team,

seems I have a strange behavior for v:page.menu.

`<v:page.menu as="navigation" pageUid="{settings.pids.primary}" entryLevel="1">`
`								<f:for each="{navigation}" as="item">`
`									<li>`
`										<f:link.typolink parameter="{item.uid}">`
`											{item.title}`
`										</f:link.typolink>`
`									</li>`
`								</f:for>`
`							</v:page.menu>`

vs. 

`page.100 = HMENU`
`page.100.entryLevel = 1`
`page.100.1 = TMENU`
`page.100.1.NO = 1`

for pages with access mode hide/show on login.
The issue is cause by my tsfe fe_user settings although not logged in.

7elix:
fixed on master already.
could be fixed with additional cast in page service for groupData['uid']

        if (TRUE === is_array($GLOBALS['TSFE']->fe_user->user)
			|| (TRUE === isset($GLOBALS['TSFE']->fe_user->groupData['uid']) && 0 < (int) $GLOBALS['TSFE']->fe_user->groupData['uid']))

7elix:
Should I send a path for master / 2.4.1?

NamelessCoder:
TER version 3.0.0 will be released soon - and it will contain all commits from `development` so that should fix the issue.

We may need the patch for the `legacy` branch though, so if you would prepare a bug fix based on and targeted at that branch, that would be appreciated!

But I don't quite see how casting to integer solves the issue *unless* it's because the data type in that variable is some type that `0 < $varWithType` is FALSE but `(integer) $varWithType` yields an integer above 0. It could be that `groupData['uid']` is actually a CSV string with commas in the front position (e.g. `,1,2`) - in which case we need to check if `GeneralUtility::trimExplode(',', $csv)` returns an array with at least one above-zero integer value, or the VH would see the page incorrectly as "not protected".

NamelessCoder:
Closed along with https://github.com/FluidTYPO3/vhs/issues/1063

PR #1063, opened 2016-06-25
[BUGFIX] v:page.menu Render pages with access

7elix:
In specific setup the TSFE->fe_user contains array config.
Ensures that pages are shown with access setup: "hide on login".

### DIFF
--- a/Classes/Service/PageSelectService.php
+++ b/Classes/Service/PageSelectService.php
@@ -65,7 +65,7 @@ class PageSelectService implements SingletonInterface {
 	 */
 	private function createPageSelectInstance() {
 		if (TRUE === is_array($GLOBALS['TSFE']->fe_user->user)
-			|| (TRUE === isset($GLOBALS['TSFE']->fe_user->groupData['uid']) && 0 < $GLOBALS['TSFE']->fe_user->groupData['uid'])) {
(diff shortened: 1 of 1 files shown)


7elix:
#1062

NamelessCoder:
Hi Felix,

I'm not sure this is an issue any longer? @bjo3rnf did a major refactoring of the menu ViewHelpers to scrap much of our custom methods to get menu etc. - I've gone through the code as it is now on the development branch, and it doesn't seem like we're selecting pages this way anymore? Maybe @bjo3rnf can elaborate!

In any case, the PageSelectService class no longer exists on the development branch so if we need this patch, we'll need it put in another place (which again I'm hoping @bjo3rnf can point out to save me some time). On a general level I recommend always creating a new branch based on development, then publishing that to your fork and creating the PR from there - makes it easier to add more commits or amend the existing one without messing with your clean state on your local development branch ;)

bjo3rnf:
> it doesn't seem like we're selecting pages this way anymore

That's correct. Current development branch should work as expected.

NamelessCoder:
I'll take the liberty of closing this one - if there still is an issue on the `development` branch we of course need to fix that, but this patch will not apply on that branch any longer.

7elix:
Not a problem with develop

PR #1067, opened 2016-07-14
[BUGFIX] Make argument optional

bjo3rnf:


NamelessCoder:
Instead of passing `false` you can omit the fourth argument and it will be optional. There probably are a *lot* of other places in current code where this happen (including passing `null` as fifth argument) - but we should aim to simplify those so as well, in other PRs ;)

bjo3rnf:
> Instead of passing false you can omit the fourth argument and it will be optional

Oh my. Of course... Btw I marked it a bugfix because I am not able to use inline notation otherwise. I will tackle those other occurrences asap.

NamelessCoder:
Bugfix is the proper type for this one since it clearly should support child content! Standing by to merge (fine to address CGL things in another PR).

PR #1074, opened 2016-07-21
[BUGFIX] Remove pages not to be shown from menu

bjo3rnf:
This patch fixes an issue where pages that are not to be shown in a menu due to translation settings still were rendered with empty link text.


bjo3rnf:
Oh the failing tests...

Issue #1089, opened 2016-08-17
v:asset.script lacks the integrity attribute


dj-rabel:
In TYPO3 8, they added the possibility to add a checksum for the HTML5 integrity attribute.
https://docs.typo3.org/typo3cms/TyposcriptReference/Setup/Page/Index.html#includejs-array
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

It would be very nice, if we could have this attribute also at least at the StyleViewHelper.
Maybe in addition, there could be a automatic generation flag so we don't have to add it manually or evan for the automatic generated vhs-assets file...

NamelessCoder:
We could implement this into the automatically generated script tags - but we could not implement it in a way that allows you to consistently define this manually in the ViewHelper itself (since assets are not defined 1:1 to `<script>` and `<link>/<style>` tags and the VH arguments all concern the asset itself, not the tag that will be used to load it).

Thanks for bringing it up :)

4-lom:
Your welcome.

I think, I've done a nice implementation of that feature allready in my fork. I'm going to test and commit it later this day.
Weather it works or not, i will write back then and eventually create a pull request for this feature.

4-lom:
Ok, it's done. Hope u will like it 👍

InvisibleKind:
I'm curious if auto-generation of an integrity hash will not make a security leak. Bcs the idea of SRI is, that browser needs to verify the provided hash, and if browser's calculation doesn't match the data provided by script-tag, the js-file is refused and not executed.

So, in case file on CDN was replaced with some credit-card-sniffer, the hash will be different.
But in case of auto-generated hash it will be generated for sniffer as well and will be correct form browser's point of view.

Please, correct me if I'm wrong.

cedricziel:
You're absolutely correct, @InvisibleKind . Yet this would take us half-way and at least make sort of a "transport-guarantee".

NamelessCoder:
The SRI should only be possible for local resources - anything else would be a security issue and if that's the case we need a fix. For remote resources you do need to specify this manually (and not use `v:asset.script` to create the tag since those don't support manual SRI for exactly that reason.

If someone would check if we're okay security wise that would be appreciated. I'm a little short on time.

NamelessCoder:
Solved, but wasn't closed along with merging the PR.

PR #1094, opened 2016-08-18
[FEATURE] Integrity attribute support for assets

4-lom:
This feature adds automatic generation of HTML5 integrity-hashes for all css and js assets included with vhs tools.
It is also possible to add the hash manually. This is useful for external scripts, were the automatic generation doesn't works.

Attibute can be true, false or a integrity-hash. true is the default behavior.

See: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
Example: '<v:asset.script path="assets/bootstrap/3.3.7/bootstrap.min.js" integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUP..." />'
Resolves: #1089

### DIFF
--- a/Classes/Asset.php
+++ b/Classes/Asset.php
@@ -63,6 +63,8 @@ use TYPO3\CMS\Extbase\Reflection\ObjectAccess;
 class Asset implements AssetInterface
 {
 
+    const INTEGRITY_METHOD = 'sha384';
+
     /**
      * @var \TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface
      */
@@ -137,6 +139,11 @@ class Asset implements AssetInterface
      * @var boolean
      */
     protected $rewrite = true;
+
+    /**
+     * @var boolean|string
+     */
+    protected $integrity = true;
 
     /**
      * @var array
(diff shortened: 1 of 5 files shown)


NamelessCoder:
To make this manageable and more importantly maintainable I would very much prefer if you would implement this so it becomes a setting you enable and if enabled, integrity is built for all tags - and leave it out of both the ViewHelper and Asset APIs.

There's a very long history of feature suggestions for our Assets which try to achieve this method of operating and they all end up significantly changing method signatures and making properties behave in multiple ways (which is the part that gets very hard to maintain) because of the underlying problem: Assets are considered disconnected from the tags that include them and every feature that tries to affect how the *tags* are built should for this reason never be configured via the Asset itself. It is very intentionally designed it this way; exactly to support the efficient grouping and merging of assets even when they have dependencies.

One example that comes to mind is the popular request to add a "media" attribute. This adds an additional layer of sorting of Assets (to group them further by their "media" attribute) - and we are sort of getting the same result here in that integrity-enabled and -disabled assets or assets with manual integrity set, need to transfer this variable (and so causes method signature changes; which if we accepted, would invite every future feature addition to do the same).

I haven't dug through the archives to find the thread(s) where this has been discussed previously but hopefully you can deduce from the above just how dangerous (in terms of maintainability and future features) it would be to implement this type of strategy with the API. If you change it to a global flag that "just" does this for all VHS assets if enabled then that's a completely different story: it gets less flexible but hugely more maintainable. And being opt-out, if as an integrator you run into issues with some assets not working (I couldn't see why but never say never! After all, IE exists) then you could still disable it (it should not be enabled by default either). Lastly, for those cases where you absolutely must be able to generate a `<script>` or other tag in the header with properties VHS can't support or you don't want to enable the global flag - `v:asset` exists to allow inserting basically anything in the header (you write the entire tag and everything in this VH).

4-lom:
Ok, thanks for the fast feedback. I think u know it much better then I do  ;-)

In this commit, I allready have a global 'opt-out' flag. I will switch it to a global 'opt-in' strategy, remove the api changes and calling back then.

Can u anwer me one more question... The Travis CI build failes are not comming from my changes, or do they? I havent seen errors from my changes in the log, but I dont really know much about Travis.

NamelessCoder:
> In this commit, I allready have a global 'opt-out' flag. I will switch it to a global 'opt-in' strategy, remove the api changes and calling back then.

Perfect! If you have any questions ask away ;)

> Can u anwer me one more question... The Travis CI build failes are not comming from my changes, or do they?

No, they appear to be caused by a change in the TYPO3 core we test against (namely the removal of a method we mock). I'll try to repair that meanwhile so your next PR will build correctly.

PR #1097, opened 2016-08-18
[FEATURE] Subresource Integrity (SRI) generation for v:asset.script …

4-lom:
…and style

Vhs can now automaticly generate and add the SRI to the resulting asset tag.
This can be turned on through the option 'assets.tagsAddIntegrityAttribute'.
These values are possible: 0 = off | 1 = sha256 | 2 = sha384 | 3 = sha512

Example: plugin.tx_vhs.assets.tagsAddIntegrityAttribute = 2

This will result in something like...

Example: `<script type="..." src="js/your.js" integrity="sha384-6XaWTzGb6esk97...">`

Resolves: #1094

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -241,6 +241,7 @@ class AssetService implements SingletonInterface
      */
     protected function buildAssetsChunk($assets)
     {
+        $setup = &$GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_vhs.'];
         $spool = [];
         foreach ($assets as $name => $asset) {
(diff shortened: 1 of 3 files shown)


NamelessCoder:
Exactly what we needed ;) thanks!

cedricziel:
Great! Thanks a bunch!

+1 from my side

NamelessCoder:
Manually picked - thanks Ruben :)

typo3ua:
This is feature only work when attribute standalone="0" and SRI Hash is generated automatically
How to be if I want to add in the integrity attribute the own value which I get from https://srihash.org/?

PR #1098, opened 2016-08-19
Add Attribute "sizes" to render apple-touch-icon

misterboe:
Example: 
```
<v:page.header.link sizes="57x57" rel="apple-touch-icon-precomposed" href="{icon57}"/>
```

https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html


xf-:
I think we had such a PR/issue before and declined it.
A lot of attributes exists, but we only support most common and other can be used via additional attributes. http://www.w3schools.com/tags/tag_link.asp
@NamelessCoder should decide

Personally i use complete long mobile, MS Tiles, Android block via <v:page.header>

bjo3rnf:
Please see #1100

PR #1109, opened 2016-09-02
[BUGFIX] Use instantiated PageRepository in PageService

pfeiferbit:
This change uses the correct instance of PageRepository within PageService.

Trying to access the TSFE within the TYPO3 backend would lead to a PHP Fatal Error because `$GLOBALS['TSFE']` is `null`.

I believe this was an oversight during the update to 3.0.0 or 3.0.1.

Best regards,
Maximilian Pfeifer


NamelessCoder:
Thanks Maximilian :)

bjo3rnf:
Oups! Thanks @pfeiferbit +1

pfeiferbit:
No problem!

PR #1112, opened 2016-09-06
[BUGFIX] Page\Resources\FalViewHelper cannot fetch resources for localised pages

dmitryd:
This viewhelper incorrectly processed page localisations by fetching localisation data directly from pages_language_overlay table. This way it ignores all kind of overlay modes and also replaces the uid of the page record. Replacing the uid breaks resource relations in localised languages or "ALL" language (sys_launguage_uid == -1). This resources are missing in any language except the default.

This commit solves the issue by correctly localising page recods via PageRepository.


dmitryd:
Only 012e175 is needed. Others come from github fork syncing. I am not sure how to get rid of them :(

NamelessCoder:
Thanks Dmitry - I will pick the change manually to exclude the other commit.

NamelessCoder:
Merged as https://github.com/FluidTYPO3/vhs/commit/2a4c6b9b9586d791706a98c6116476e3ebf67ea1

Issue #1115, opened 2016-09-13
 Call to undefined method FluidTYPO3\Vhs\ViewHelpers\Page\Menu\SubViewHelper::isActive() 


Danny-P:
Hi,

I get the following error: 
Fatal error: Call to undefined method FluidTYPO3\Vhs\ViewHelpers\Page\Menu\SubViewHelper::isActive() in /srv/www/typo3/htdocs/typo3conf/ext/vhs/Classes/ViewHelpers/Menu/SubViewHelper.php on line 47

I'm using Typo3 7.6.10 and vhs 3.0.1

PR #1122, opened 2016-09-20
[BUGFIX] Replace call to superfluous method with service call

tiggr:
refs: #1115, #1116


cedricziel:
Seems i missed that one! Nice catch, thank you for your contribution! :)

PR #1126, opened 2016-10-01
[BUGFIX] Fix parameter usage in v:page.header.canonical

JostBaron:
Before this patch, the ViewHelper ignored parameters set by CoolUri (and maybe Realurl as well).

Now it takes parameters either from `GeneralUtility::_GET()`, `GeneralUtility::_POST()` or both, actually adding those parameters to the generated URL.


NamelessCoder:
I just noticed that this patch was made to the `master` branch instead of the correct `development` branch. In the future please always target the `development` branch. Coming release 3.1.0 will bring it up to date.

Issue #1103, opened 2016-08-25
slide not working on FluidTYPO3\Vhs\ViewHelpers\Page\Resources\FalViewHelper


educo-adymorz:
I try to display the first media element on page using slide -1. On the page the record is assigned to it works fine, put on any descendant page no resources are fetched.

``` html
<v:page.resources.fal table="pages" field="media" as="resources" slide="-1">
    <v:variable.set name="bgImage" value="{resources.0.url}"/>
</v:page.resources.fal>
<if condition="{bgImage}">
    <v:asset.style content=".image-bg { background-image: url('/{bgImage}'); }" />
</if>
```

It worked in vhs 2.4.0, but does not work in vhs 3.0.1

Edit: I am using TYPO3 CMS 7.6.10

jobee:
@educo-adymorz same here... have you found a solution?

educo-adymorz:
@jobee Unfortunately did not find the time to have a closer look yet. Issue still exists.

educo-adymorz:
To be able to debug I added the render method to FluidTYPO3\Vhs\ViewHelpers\Page\Resources\FalViewHelper as an entry point for debugging.

```php
    public function render()
    {
        return parent::render();
    }
```

While debugging I found out, that the FalViewHelper render method gets called twice, although I only call it once. On the first run it returns zero resource records, on the second it returns one as expected.

I stop debugging for now because it took me more time than expected :-(

educo-adymorz:
Because I did not find the cause of the problem I decided to go for a typoscript workaround:

```
lib.backgroundImage = FILES
lib.backgroundImage {
	references {
		data = levelmedia:-1,slide
	}
	renderObj = TEXT
	renderObj {
		data = file:current:publicUrl
        wrap (
            <style type="text/css">
                .image-bg { background-image: url('/|'); }
            </style>
        )
	}
}

page.headerData.201609131646 < lib.backgroundImage
```

cedricziel:
That's a solid solution to the problem. Thanks for getting back to it

ogrosko:
Hey I have the same problem at Typo3 v 7.6.11 and vhs 3.0.1
Is there a change to be fixed? or should I downgrade vhs to v 2.4.0

Thanks

mkrappitz:
JFYI: As with the current vhs version 4.2.0 and t3 7.6.21 this does not work:
<v:page.resources.fal table="pages" field="media" as="resources" slide="-1">
...
but this works:
[...]

PR #1127, opened 2016-10-05
[BUGFIX] Slides now working. Missing call to getSlideRecords.

monosize:
Change getSlideRecordsFromPage. to only check if there are valid resources.
#1103

### DIFF
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -11,6 +11,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Page\Resources;
 use FluidTYPO3\Vhs\ViewHelpers\Resource\Record\FalViewHelper as ResourcesFalViewHelper;
 use FluidTYPO3\Vhs\Traits\SlideViewHelperTrait;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Fluid\Core\ViewHelper\Exception;
 use TYPO3\CMS\Frontend\Page\PageRepository;
 
 /**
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hey @monosize - could you give a quick explanation why it is necessary to duplicate the entire `render()` method? The functionality it provides, including the sliding, should already be provided by either base class or Trait.

monosize:
@NamelessCoder - in the render method from Class ```FluidTYPO3\Vhs\ViewHelpers\Resource\Record\AbstractRecordResourceViewHelper```  is no call to ```getSlideRecords```. 
see in AbstractRecordResourceViewHelper https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Resource/Record/AbstractRecordResourceViewHelper.php#L197

and here in the PR https://github.com/FluidTYPO3/vhs/pull/1127/commits/0a42f58c7ca7e1b52ef4b9cb1d3108874aea6dfd#diff-3b954cee31eae4dcef6f278f8bde4461R79

NamelessCoder:
The intention is that sliding should only be available if the VH also implements the `SlideViewHelperTrait` which is what defines the `getSlideRecordsFromPage` that you are actually addressing. It looks like incorrect use of the logic contains within that Trait.

I don't have the full picture but looks to me like the proper solution would be to implement `SlideViewHelperTrait` in the `FalViewHelper`, assuming that the problem you face isn't because you actually meant to use for example `v:page.resources.fal` which is a specific FAL "record resource" ViewHelper that enables sliding because `pages` contains sliding. This record resource ViewHelper is after all intended to be generic for any record type (not just those supporting parent->child relations!). If I am missing a use case here, please elaborate ;)

Maybe this solves your use case?

NamelessCoder:
@monosize Did https://github.com/FluidTYPO3/vhs/pull/1186 also solve this issue? Seems they touch the same code at least.

NamelessCoder:
Quite sure this is solved - during my testing over the past weeks I have been using sliding the whole time.

PR #1135, opened 2016-10-19
fix language fallback

dhelfer:


dhelfer:
Yes you're right.
So the better approach would be to use the defined content fallback from config.sys_language_mode.
https://docs.typo3.org/typo3cms/TyposcriptReference/Setup/Config/Index.html#sys-language-mode
Example: config.sys_language_mode = content_fallback; 1,0 --> fill -1, 0, 1 and currentlanguage into this incondition. Without a fallback defined just -1 and currentlanguage.

dhelfer:
would this be a suitable adaption?

NamelessCoder:
> would this be a suitable adaption?

Yes, assuming you put the selected language first and we don't risk getting multiple results from the query, that should be okay.

jacobsenj:
I think a different approach is needed to support fallback languages for the content viewhelpers. The whole topic is not an easy one. I would opt to remove the added code to get base stuff working and tackle the content fallback in a new run.

NamelessCoder:
Language handling is now delegated to the core's record fetching method via ContentObjectRenderer. Thanks for your patience guys!

PR #1136, opened 2016-10-19
[DOC] Update changelog for 3.1.0

InvisibleKind:


NamelessCoder:
Thanks Viktor :)

I actually forgot to do this for the 3.1.0 and then at long last decided to create scripts that will do this for us from now on. The build tools will now write the change log every time a release is created simply as part of the release process.

Your work on these change logs has been truly awesome, can't thank you enough for doing this manually for so long. In a while I will be converting our change logs to the new approach so it all can happen automatically in the future.

NamelessCoder:
If you are curious you can see an example of the new automatic logs at https://github.com/FluidTYPO3/temp-releaseflow. Briefly described, it does a nice `git log` and updates the main CHANGELOG.md as well, listing most recent version and all previous versions below in descending order.

cedricziel:
Thank you, Viktor!

InvisibleKind:
Oh no! This day has come: I'm replaced with a robot...

But seriously, thank you guys for the awesome set of tools, so I was so inspired to put my hands onto changelogs :)
I'm not leaving anyway. My next step would be to participate into some bugfixes or feature requests, if I'll be brave enough)

NamelessCoder:
> My next step would be to participate into some bugfixes or feature requests, if I'll be brave enough)

Please do! We would appreciate that a lot :)

PR #1139, opened 2016-10-19
[TASK] Update coding style

NamelessCoder:
* PSR-2 reapplied
* Shorthand array syntax converted
* Class name string values replaced with static references

### DIFF
--- a/Tests/Unit/ViewHelpers/Iterator/ExtractViewHelperTest.php
+++ b/Tests/Unit/ViewHelpers/Iterator/ExtractViewHelperTest.php
@@ -9,6 +9,7 @@ namespace FluidTYPO3\Vhs\Tests\Unit\ViewHelpers\Iterator;
  */
 
 use FluidTYPO3\Vhs\Tests\Unit\ViewHelpers\AbstractViewHelperTest;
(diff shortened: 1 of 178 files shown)


xf-:
Nice work, some doc headers for resolving in IDE would be great. But i think its not so important move step by step forward :)

NamelessCoder:
Ready to be merged

Tests/Unit/ViewHelpers/Iterator/ExtractViewHelperTest.php
```diff
@@ -30,7 +31,7 @@ class ExtractViewHelperTest extends AbstractViewHelperTest
     public function setUp()
     {
         parent::setUp();
-        $this->fixture = $this->getAccessibleMock('FluidTYPO3\Vhs\ViewHelpers\Iterator\ExtractViewHelper', array('hasArgument'));
+        $this->fixture = $this->getMockBuilder(ExtractViewHelper::class)->setMethods(['hasArgument'])->getMock();
```
mneuhaus: Please keep "Update coding style" Tasks to styling in the future ;) it's ok i guess, but makes things like this easier to review if its really only styling updates

PR #1142, opened 2016-10-23
[REMOVAL] Deprecate v:if in favor of f:if

NamelessCoder:
Deprecate and plan to remove `v:if` (which has the
stack condition argument) in favor of `f:if` as provided
by Fluid Standalone (which parses complex expressions).

To be removed when TYPO3v8 LTS is our minimum.

NB: deprecation logged from initializeArguments in order
to limit the number of logged deprecations to one per
request where v:if is used rather than once per execution.


NamelessCoder:
The `[REMOVAL]` prefix is revived to be used instead of `[TASK]` when the task is one that removes or deprecates code. This prefix will get the commit included in the automatic change log.

PR #1149, opened 2016-11-06
[FEATURE] Add extension configuration toggle to disable Assets

NamelessCoder:
Since Assets are the only part of VHS which cause runtime
resource usage regardless of whether you use it or not, a
toggle is added which allows disabling the Asset handling.

Doing so will not break the ViewHelpers - VHS will simply
stop processing the ViewHelpers' generated assets.

### DIFF
--- a/Classes/ViewHelpers/Asset/AbstractAssetViewHelper.php
+++ b/Classes/ViewHelpers/Asset/AbstractAssetViewHelper.php
@@ -235,7 +235,11 @@ abstract class AbstractAssetViewHelper extends AbstractViewHelper implements Ass
      */
     public function render()
     {
-        $this->finalize();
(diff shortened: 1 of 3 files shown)


mneuhaus:
Could you elaborate, why this should be put in? Does it serve an edgecase or improve performance?

NamelessCoder:
It serves as a performance improvement when nothing on the site uses assets. I had also considered making it an opt-in feature to avoid performance issues.

The problem is that if the FE is extremely fast in itself (<50ms) the VHS Asset hook begins consuming +10% wall time because it needs to read TS (made worse by the fact there are two hooks).

Even triggering the hooks consumes the memory and wall time also without assets. Which isn't good, I agree, but a toggle-off can't be beaten in terms of performance :)

PR #1150, opened 2016-11-07
Bugfix: Backup TSFE to reset it later when simulating frontend environment

Lumrenion:
When in development mode, the TYPO3 backend showed PHP warnings, because the argument for method resetFrontendEnvironment() was missing.


NamelessCoder:
We plan to go a compltely different way - see https://forge.typo3.org/issues/73183

mneuhaus:
can this be closed then?

NamelessCoder:
Maybe we should do this as a temporary fix until that decision is made. If you do look into that please make sure the commit CGL is right before squashing ;)

PR #1151, opened 2016-11-12
[TASK] Preparations for LTS-v8 compiling support bridge

NamelessCoder:
This patch adds a couple of changes which will be required
for an in-progress conversion of every ViewHelper which
can support it, to ideal performance when compiled.

A new dependency, namelesscoder/typo3-cms-fluid-gap,
provides the necessary API to achieve this. Slight rewriting
of the test base ensures all ViewHelpers can be tested with
proxying from render() to renderStatic().

### DIFF
--- a/Classes/Utility/ViewHelperUtility.php
+++ b/Classes/Utility/ViewHelperUtility.php
@@ -1,0 +1,34 @@
+<?php
+namespace FluidTYPO3\Vhs\Utility;
+
+/*
+ * This file is part of the FluidTYPO3/Vhs project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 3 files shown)


Classes/Utility/ViewHelperUtility.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Fixes a bug in TYPO3 6.2.0 that the properties metadata is not overlayed on localization.
+     *
+     * @param RenderingContextInterface $renderingContext
+     * @return TemplateVariableContainer|\TYPO3Fluid\Fluid\Core\Variables\VariableProviderInterface
+     */
+    public static function getVariableProviderFromRenderingContext(RenderingContextInterface $renderingContext)
+    {
+        if (method_exists($renderingContext, 'getVariableProvider')) {
```
mneuhaus: return early? ;)

NamelessCoder: Yeah, why not :)

PR #1153, opened 2016-11-12
[TASK] Implement compiling bridge in Variable scope

NamelessCoder:


Classes/ViewHelpers/Variable/TyposcriptViewHelper.php
```diff
@@ -44,36 +46,51 @@
...
+    protected static $configurationManager;
 
     /**
      * @param ConfigurationManagerInterface $configurationManager
      * @return void
      */
     public function injectConfigurationManager(ConfigurationManagerInterface $configurationManager)
     {
-        $this->configurationManager = $configurationManager;
+        static::$configurationManager = $configurationManager;
```
mneuhaus: a static instance of the configurationManager? could this cause context issues?

NamelessCoder: The edge case that might be affected is if a sub-rendering in Fluid overrides the configurationmanager implementation during runtime. It's so remote I don't worry about it - normally, the configuration manager is the same across the entire request.

Issue #1075, opened 2016-07-21
v:render.uncache doesn't respect partialRootPaths


smichaelsen:
Though `v:render.uncache` should be usable just like `f:render` there's an important difference: When you rely on non-standard `partialRootPaths` your partials won't be found/resolved correctly.

That's because the "uncached" rendering is done with a new view object (`UncacheTemplateView`), which doesn't get the path settings from the view object that is rendering the rest of the templates.

Immediately, I didn't get a good idea how to pass those settings properly - but maybe someone else does.

smichaelsen:
I implemented a fix for myself. Don't know if this is a beautiful solution though.
https://gist.github.com/smichaelsen/6d76c7afaf050741441dd083f6e29939

typo3ua:
When I use this Helper <v:render.uncache/> get the error:

=My code:
``
<v:render.uncache partial="Subscription/Form" arguments="{_all}"/>
``
=Error:
``
Oops, an error occurred!
TYPO3 Fatal Error: Extension key "" is NOT loaded!
More information regarding this error might be available online.
``
Why?

typo3ua:
I found this [solution ...](https://www.stephan-meier.com/vhsrender-uncache-viewhelper-typo3-fatal-error/)

NamelessCoder:
Solved by https://github.com/FluidTYPO3/vhs/commit/530649c52bc5629ecae7a22b428c7c09a17a3769

smichaelsen:
@NamelessCoder I guess #1163 should also be closed then.

PR #1163, opened 2016-11-13
[BUGFIX] RenderUncache ViewHelper

monosize:
using the correct templateRootPaths, layoutRootPaths and partialRootPaths

### DIFF
--- a/Classes/View/UncacheTemplateView.php
+++ b/Classes/View/UncacheTemplateView.php
@@ -10,7 +10,10 @@ namespace FluidTYPO3\Vhs\View;
 
 use TYPO3\CMS\Core\Cache\CacheManager;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 use TYPO3\CMS\Extbase\Mvc\Controller\ControllerContext;
+use TYPO3\CMS\Extbase\Mvc\View\ViewInterface;
+use TYPO3\CMS\Extbase\Utility\ArrayUtility;
 use TYPO3\CMS\Fluid\Compatibility\TemplateParserBuilder;
 use TYPO3\CMS\Fluid\Core\Compiler\TemplateCompiler;
 use TYPO3\CMS\Fluid\Core\Rendering\RenderingContext;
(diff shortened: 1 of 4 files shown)


Classes/View/UncacheTemplateView.php
```diff
@@ -67,6 +73,76 @@ protected function prepareContextsForUncachedRendering(
     }
 
     /**
+     * @param ViewInterface $view
+     *
+     * @return void
+     */
+    protected function setViewConfiguration(ViewInterface $view)
```
NamelessCoder: I'm afraid we can't accept this huge chunk of View initialisation, for a couple of reasons:

* It is duplication of methods already existing on controllers
* The information about paths is already present in the view
* The patch will be calling deprecated methods on v8+
* The applied reverse sorting will break template resolving on v8+

But it does look like we may need to copy a bit more info than just controller context. I suggest extracting the template paths (note: on v8+ this is done with `$renderingContext->getTemplatePaths()->toArray()`) and then add those paths to the serialised configuration we put in the USER_INT init and restore them (using `$renderingContext->getTemplatePaths()->fillFromConfigurationArray($thawedPaths)` on v8+, e.g. when the method `getTemplatePaths` exists on rendering context when reading context, or if `fillFromConfigurationArray` exists when initialising the uncached view. Otherwise, call the "old school" methods on View to setTemplateRootPaths etc.

monosize: Sorry, the information about the paths are definitively not present in the view (TYPO3 7.6.x).
In the view are only the calculated paths (one for layout, partial and template), based from the extension name. 

The paths defined in TS ( view.templateRootPaths ...) are not respected.

If have spent some time in debugging to find out the problem.

cedricziel:
How do we go on with this one?

bjo3rnf:
Related to #1075

NamelessCoder:
This could easily indicate a bug in TYPO3 itself. The view should most definitely contain the paths we need - and the (Widget-)controller should also be aware of the various context pieces so it can initialise the View correctly. These should reside in the ControllerContext on 7.6 and TemplatePaths on v8+ so if they are not filled there, it indicates a problem further up, i.e. in the controller which rendered the template.

I should also note that when using FLUIDTEMPLATE it is expected that the paths are present as actual paths arrays (as opposed to just knowing the extension key and then being able to read the TS configured paths).

I of course don't mean to disregard the effort made by @monosize on this one, but I am completely convinced that reproducing all of this core business logic should not be necessary. if the View doesn't know about either extension scope or defined paths, that's a problem not just for this ViewHelper but for other use cases as well (e.g. any Widget).

NamelessCoder:
Replaced by https://github.com/FluidTYPO3/vhs/commit/530649c52bc5629ecae7a22b428c7c09a17a3769 - we've gone with a much more limited solution that only transfers the partial root paths since layout- and template root paths don't make much sense in this context. Even though it didn't get merged: thank you @monosize for your work!

PR #1170, opened 2016-11-14
[TASK] Implement compile bridge in Iterator scope

NamelessCoder:
All except for ValuesViewHelper which is unfortunately
not sufficiently covered by tests yet.

### DIFF
--- a/Classes/ViewHelpers/Iterator/AbstractLoopViewHelper.php
+++ b/Classes/ViewHelpers/Iterator/AbstractLoopViewHelper.php
@@ -8,6 +8,8 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Iterator;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use FluidTYPO3\Vhs\Utility\ViewHelperUtility;
+use TYPO3\CMS\Fluid\Core\Rendering\RenderingContextInterface;
 use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper;
 
 /**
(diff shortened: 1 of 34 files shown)


cedricziel:
Just for my sanity: Do we package the fluidgap package on build and TER upload?

NamelessCoder:
We do not - and I see where you're going. Any ways we could bundle this somehow and have it 1) only be loaded when TYPO3 is in non-composer mode and 2) not conflict with potential other extensions doing the same thing?

cedricziel:
Hmm. The best we could do is to require the polyfill as both extension and composer package so we can guarantee it's available. What do you think?

cedricziel:
oh: important sidenote, you will have guessed it; - this would require releasing the polyfill as versioned extension.

NamelessCoder:
In that case I think I'd rather cheat and include the two traits in Resources/Private/Php/FluidGap/Traits and only load them if TYPO3 is not in composer mode and the classes do not already exist, or alias the existing traits if TYPO3 is 8.4+.

Just the thought of having to maintain a TER extension for two traits or two aliases, and do so in two major version branches, makes me shiver a bit ;)

cedricziel:
Hmm. Yeah, that's quick and dirty. But mostly for the sake of having two different "meanings" for a package. ;)

cedricziel:
How do we keep going with these?

NamelessCoder:
Plan to continue:

* Set minimum requirement to TYPO3 7.6.13
* Convert current usages of Traits to point directly to Fluid Standalone's Trait name
* Convert the rest of the lib

See https://review.typo3.org/#/c/50676/

NamelessCoder:
See also https://review.typo3.org/#/c/50714/ - depending on the decision there, VHS ViewHelpers either must or should not implement CompilableInterface. Un-friggin-fortunately because PHP Traits, we can't have the Trait implement that interface (which would be the clean solution for the backwards compatible stuff).

bjo3rnf:
Is there some general explanation what this is all about and does this mean vhs now has a dependency?

NamelessCoder:
> Is there some general explanation what this is all about 

The short version is to allow ViewHelpers to compile on 7.6 and 8.4+, which isn't natively possible because of differences in method signatures (on the compile() method).

> and does this mean vhs now has a dependency?

No, not any more - since 7.6.13 we now have some backwards compatible Traits which use the same name on 7.6 and 8.4, so we no longer need the cms-fluid-gap dependency; which was always intended to be temporary - but now is entirely redundant.

bjo3rnf:
Ok, I see. Thanks.

cedricziel:
What's missing here - a rebase?

NamelessCoder:
Missing a class import - fixed :)

cedricziel:
Thx!

PR #1171, opened 2016-11-14
[TASK] Bundle FluidGap classes for 7.6 in non-composer mode

NamelessCoder:
Uses a bit of careful decision making before loading the
classes, since it is more than likely other extensions might
be using the same trick to include these non-typo3-extension
classes on non-composer setups.

### DIFF
--- a/Resources/Private/Php/FluidGap/CompileWithContentArgumentAndRenderStatic.php
+++ b/Resources/Private/Php/FluidGap/CompileWithContentArgumentAndRenderStatic.php
@@ -1,0 +1,109 @@
+<?php
+namespace NamelessCoder\FluidGap\Traits;
+
+use TYPO3\CMS\Fluid\Core\Compiler\TemplateCompiler;
+use TYPO3\CMS\Fluid\Core\Parser\SyntaxTree\AbstractNode;
+use TYPO3\CMS\Fluid\Core\ViewHelper\Exception;
+
+
+/**
+ * Class CompilableWithContentArgumentAndRenderStatic
+ *
+ * Provides default methods for rendering and compiling
+ * any ViewHelper that conforms to the `renderStatic`
+ * method pattern but has the added common use case that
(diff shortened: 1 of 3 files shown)


NamelessCoder:
If someone could test this with a site that has not been touched by composer, that would be appreciated. At this point I barely know how to force it to not use composer...

cedricziel:
From my experience this will fall short when in backend context, and you would need to add it to ext_tables as well.

But I would prefer it, if you added those to the autoload-maps in both composer.json and ext_emconf. We don't need to alias those classes as we can just add more psr-4 roots at will. I think classloading won't be an issue since it's first found, first require'd. Also.. as I think about it: if we vendor those in, we could vendor them in by changing their namespace and drop aliasing-since it's just two traits as you said.

NamelessCoder:
I've had some issues with composer classloading in authorative mode (`-a`) when it detects multiple presences of the same class, which would potentially happen with this. Also, those must only load exactly when the site is *not* in composer mode, so we can't really use the autoloading.

If we include these in VHS we still have a problem that they must be different according to the TYPO3 version; with 8.4+ using an alias instead.

But you're probably right about the BE context thing so that has to be considered.

NamelessCoder:
It's not pretty... but it should work. Besides, it goes away completely as soon as we discontinue 7.6, at which point there will be zero expected maintenance in regards to this little hack (it will simply go into the `legacy` branch and be removed in `development`).

NamelessCoder:
Not good! It seems TYPO3 will not be able to support this method when *not* in composer mode. When the site isn't in composer mode it will scan all classes and add all of them to the class loader.

Reference: http://stackoverflow.com/questions/40551363/composer-autoloads-classes-of-typo3-extension-without-any-autoload-configuration

Will have to find another method. Ach!

NamelessCoder:
Good news, everyone!

https://review.typo3.org/#/c/50676/2

PR #1174, opened 2016-11-22
Support ObjectStorages on SelectOptionViewHelper in case the select i…

Schweriner:
…s configured as "multiple"

Usage example:

```
<v:form.select property="kategorie" id="kategorie" multiple="1" class="form-control" optionLabelField="name" optionValueField="uid" additionalAttributes="{required: 'required'}">
                <v:form.select.option value="">
                    Bitte wählen
                </v:form.select.option>
                <f:for each="{kategorien}" as="ueberKategorie">
                    <v:form.select.optgroup label="{ueberKategorie.name}">
                        <f:for each="{ueberKategorie.unterkategorien}" as="kategorie">
                            <v:form.select.option value="{kategorie.uid}">
                                {kategorie.name}
                            </v:form.select.option>
                        </f:for>
                    </v:form.select.optgroup>
                </f:for>
            </v:form.select>
```

### DIFF
--- a/Classes/ViewHelpers/Form/Select/OptionViewHelper.php
+++ b/Classes/ViewHelpers/Form/Select/OptionViewHelper.php
@@ -9,6 +9,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Form\Select;
  */
 
 use FluidTYPO3\Vhs\ViewHelpers\Form\SelectViewHelper;
+use TYPO3\CMS\Extbase\Persistence\ObjectStorage;
 use TYPO3\CMS\Fluid\ViewHelpers\Form\AbstractFormFieldViewHelper;
 
 /**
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Form/Select/OptionViewHelper.php
```diff
@@ -57,6 +57,15 @@ public function render()
             if (false === is_object($this->arguments['value']) && false === is_array($this->arguments['value'])) {
                 if (true === is_array($value)) {
                     $selected = true === in_array($this->arguments['value'], $value) ? 'selected' : '';
+                } else if($value instanceof \TYPO3\CMS\Extbase\Persistence\ObjectStorage && is_numeric($this->arguments['value'])) {
```
cedricziel: Please import the class instead of using the FQCN

Classes/ViewHelpers/Form/Select/OptionViewHelper.php
```diff
@@ -57,6 +57,15 @@ public function render()
...
+                    // Requires that the option values are UIDs of objects in ObjectStorage
+                    $objectStorageObjects = $value->toArray();
```
cedricziel: you can directly loop on an objectstorage without explicitly calling `toArray` since it implements `ArrayAccess`.

Classes/ViewHelpers/Form/Select/OptionViewHelper.php
```diff
@@ -57,6 +57,15 @@ public function render()
...
+                    foreach ($objectStorageObjects as $object) {
+                        if($object->getUid() == $this->arguments['value']) {
```
cedricziel: Please use strict comparisons

Classes/ViewHelpers/Form/Select/OptionViewHelper.php
Schweriner: No idea what I though of not doing it like that immedately.

Schweriner:
Should be fine, now.

cedricziel:

cedricziel:
Okay, my fault-had to force push, since I forgot about our CGL. You might want to reset your local branch if you already pulled.

PR #1175, opened 2016-11-25
[FEATURE] Add Iterator/Split ViewHelper for str_split

NamelessCoder:
Provides a way to split a string into an array with $length
number of bytes of the string in each array value.

In other words, `str_split` for Fluid.


NamelessCoder:
Thanks @bjo3rnf :)

PR #1189, opened 2016-12-09
TASK: Change tablenames in query to sys_file_metadata

pide2000:
The correct tablename is sys_file_metadata if one use categories for files. Alternatively we can use the fieldname which is "categories".  So this whole thing will be changed to fieldname = categories. 
Discussion is very welcome


NamelessCoder:
Have you confirmed this one working on 7.6 LTS without EXT:filemetadata installed? Correct me if I'm wrong, but `sys_file_metadata` is added by this extension on 7.6 (and is added by the core itself on 8.x). If that's the case we at least need a check around this before using the table, and possibly we need a check both for installed TYPO3 version as well as installed state of EXT:filemetadata.

pide2000:
Hey, thank you for the hints. That's totally true. I will fix the pull request.

NamelessCoder:
Thanks - you've got a few more days before we plan to do the next release ;)

PR #1198, opened 2016-12-28
[BUGFIX] Fix incorrect content fetching in workspaces

NamelessCoder:
This patch fixes all content-related ViewHelpers when
used in combination with workspaces. Solves issues
which caused live and preview workspaces to show
incorrect content or wrongly positioned content.

Fix kindly sponsored by MSI Chicago - http://www.msichicago.org/

### DIFF
--- a/Classes/ViewHelpers/Content/AbstractContentViewHelper.php
+++ b/Classes/ViewHelpers/Content/AbstractContentViewHelper.php
@@ -9,6 +9,8 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Content;
  */
 
 use FluidTYPO3\Vhs\Traits\SlideViewHelperTrait;
+use TYPO3\CMS\Backend\Utility\BackendUtility;
+use TYPO3\CMS\Core\Utility\ExtensionManagementUtility;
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Content/AbstractContentViewHelper.php
```diff
@@ -140,15 +142,30 @@ protected function getSlideRecordsFromPage($pageUid, $limit)
...
         }
-        $conditions .= ' AND t3ver_state <= 1';
+        if (ExtensionManagementUtility::isLoaded('workspaces')) {
+            $conditions .= BackendUtility::versioningPlaceholderClause('tt_content');
+        }
 
         $rows = $GLOBALS['TYPO3_DB']->exec_SELECTgetRows('*', 'tt_content', $conditions, '', $order, $limit);
+
+        if (ExtensionManagementUtility::isLoaded('workspaces')) {
+            $workspaceId = isset($GLOBALS['BE_USER']) ? $GLOBALS['BE_USER']->workspace : 0;
```
mneuhaus: could be switched to a return early ;D

NamelessCoder: Switched :)

PR #1200, opened 2017-01-03
[TASK] Loosen page record constraints

### LINKED ISSUE #1114: empty array from page.info of sysfolder
Dear vhs-team,

thank you for your great work. 
After updatinig the typo3 extension from V 2.4.0 to 3.0.1 the page.info viewhelper does not act as before: 
If you call {v:page.info(pageUid: '17')}  and the PID is from a sysfolder, the output array ist empty. If you change pageUid to eg. 1, an array of the home page is given. I changed back to 2.4.0 and the page.info gives array of sysfolder.

Typo3 7.10

regards
seebold

bjo3rnf:
This is a really pragmatic approach to fix #1114. Please be aware that no access restrictions apply here. I'm still rather for removing this viewhelper completely as we want to simplify things with vhs but try to stay as close to the core as possible. Developers who rely on this should IMHO create their own helpers to fit their use cases.

### DIFF
--- a/Classes/ViewHelpers/Page/InfoViewHelper.php
+++ b/Classes/ViewHelpers/Page/InfoViewHelper.php
@@ -8,12 +8,9 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Page;
  * LICENSE.md file that was distributed with this source code.
  */
 
-use FluidTYPO3\Vhs\Service\PageService;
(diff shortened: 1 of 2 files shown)


bjo3rnf:
Ok. How can I run a local unit test under 7.6.15? `composer require "typo3/cms:>=7.6.13, <7.6.99"` yields a dependency problem for me. I looked that up from `.travis.yml`.

Issue #1121, opened 2016-09-15
v:unless throws '#1237823695: Argument "condition" was not registered.'


InvisibleKind:
Trying to use `v:unless` but it results in an exception `#1237823695: Argument "condition" was not registered.`.

I added the following code (just like in `f:if` ViewHelper ) into UnlessViewHelper and it works again:

```
/**
 * @return void
 */
public function initializeArguments()
{
    $this->registerArgument('condition', 'boolean', 'Condition to be evaluated.');
}
```

The question is: should this code be added directly into `UnlessViewHelper` or it should inherit from `IfViewHelper`?

TYPO3 7.6.11
vhs 3.0.1

NamelessCoder:
Looks like this condition ViewHelper was missed. It should inherit from AbstractConditionViewHelper and should implement evaluateCondition() as the only overridden method - that should work.

InvisibleKind:
It actually inherits from [AbstractConditionViewHelper](https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/UnlessViewHelper.php#L47), but as I understand Fluid throws an exception, because it can't find "condition" in VH's signature. And the exception is thrown before `evaluateCondition()` could be called.

bjo3rnf:
@InvisibleKind is right. The argument `condition` is never registered. Mind making a pull request out of your solution?

PR #1201, opened 2017-01-04
[BUGFIX] Add argument `condition` to UnlessViewHelper

cedricziel:
Fixes: #1121

### DIFF
--- a/Classes/ViewHelpers/UnlessViewHelper.php
+++ b/Classes/ViewHelpers/UnlessViewHelper.php
@@ -46,6 +46,20 @@ use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractConditionViewHelper;
  */
 class UnlessViewHelper extends AbstractConditionViewHelper
 {
+    /**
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/UnlessViewHelper.php
```diff
@@ -46,6 +46,17 @@
  */
 class UnlessViewHelper extends AbstractConditionViewHelper
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function initializeArguments()
+    {
+        parent::initializeArguments();
```
NamelessCoder: If you do "parent::initializeArguments" here, argument `else` will be added which we do not want happening. The right solution for this particular VH would be to not call parent argument initialisation, but instead manually register only the "then" and "condition" arguments.

cedricziel: Addressed. Thx for the hint.

NamelessCoder:
Code comment is really helpful, thanks!

PR #1203, opened 2017-01-04
[BUGFIX] Make path site relative as specified

### LINKED ISSUE #1093: <v:extension.path.resources /> doesn't work like expected
Hey, I've found something weird ...

The description of the <v:extension.path.resources/> ViewHelper says "Site Relative path to [...]". But what you get from this ViewHelper is a path, relative to the TYPO3_MainDir. So I don't know, weather this is the correct behaviour for this ViewHelper, then the documentation should not say "Site Relative". Else we should change the line
`return ExtensionManagementUtility::extRelPath($extensionKey) . 'Resources/Public/' . $path;`
to
`return ExtensionManagementUtility::siteRelPath($extensionKey) . 'Resources/Public/' . $path;`

bjo3rnf:
Fixes #1093


NamelessCoder:
hemingwaydays-300x217

10 of 10 Hemingways approve!

bjo3rnf:
I'm overwhelmed by the sheer amount of approval.

PR #1207, opened 2017-01-16
Some general cleanups and quick findings

cedricziel:


Classes/ViewHelpers/Format/Placeholder/LipsumViewHelper.php
```diff
@@ -58,7 +58,7 @@ public function initializeArguments()
      */
     public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
     {
-        $lipsum = $argguments['lipsump'];
+        $lipsum = $arguments['lipsump'];
```
NamelessCoder: While you're at it, could you fix the typo `lipsump` -> `lipsum` here as well? (oh boy, that was broken bad...)

cedricziel: Yeah, i suspect some hop involved..

NamelessCoder: I don't know about hop, but there was definitely some tiredness going on - I think I did this one after converting a lot of other ones. And of course the unit tests don't test for this (which is also my bad hehe).

cedricziel: I think we should really investigate functional or general blackbox tests. That could save us from __lots__ of pain easily.

NamelessCoder:
See testing strategy and general progress on decoupled VH package: https://github.com/NamelessCoder/fluid-viewhelpers/tree/master/src/ViewHelpers

I found a sensible way to write functional tests for Fluid ViewHelpers, but I'm not sure it works as well for TYPO3, and much less sure it works for 7.6 LTS (without a custom View implementation).

When we move to next TYPO3 LTS maybe we can implement this solution in VHS (after we remove all the moved ViewHelpers, declare typo3/fluid-viewhelpers as dependency and officially stop making TER releases or TER becomes composer compatible).

cedricziel:
Thx for the dance! :man_dancing:

PR #1210, opened 2017-01-16
[BUGFIX] Use namespaces from arguments and pass required rendering context

cedricziel:
…ntext


Classes/ViewHelpers/Render/InlineViewHelper.php
```diff
@@ -51,7 +51,7 @@ public static function renderStatic(
         RenderingContextInterface $renderingContext
     ) {
         $content = $renderChildrenClosure();
-        $namespaces = static::getPreparedNamespaces($arguments['namespaces']);
```
cedricziel: Good catch. - Thanks!

PR #1212, opened 2017-01-19
just inverts $invert instead of doing all the magic

torvitas:


cedricziel:
Thank you for your contribution, @torvitas !

PR #1214, opened 2017-01-24
Fix release date

cweiske:
According to CHANGELOG.md the release date is wrong.


cedricziel:
Oh boy. You've got an eye for details, but the release date is technically the date of the tag in git which happened at 2016-10-18 (https://github.com/FluidTYPO3/vhs/tags).

I will add a followup to fix the CHANGELOG.md. Thank you for your contribution!

PR #1216, opened 2017-01-26
[BUGFIX] Static injection of configurationManager failed

monosize:
Injection is only done during instanciating with a call on ObjectManager->get(), but renderStatic is a static call, so no injection is done.


cedricziel:
Hey let me just merge.. ah. nevermind. ^^

Thank you!

PR #1226, opened 2017-02-08
[BUGFIX] IteratorViewHelpers - insert missing $escapeChildren for TYPO3v8+

monosize:
Inserted where necessary.
Remove $escapeOutput if set in parent classes.

### DIFF
--- a/Classes/ViewHelpers/Iterator/ImplodeViewHelper.php
+++ b/Classes/ViewHelpers/Iterator/ImplodeViewHelper.php
@@ -26,11 +26,6 @@ class ImplodeViewHelper extends ExplodeViewHelper implements CompilableInterface
     protected static $method = 'implode';
 
     /**
-     * @var boolean
-     */
-    protected $escapeOutput = false;
-
-    /**
      * Initialize
      *
      * @return void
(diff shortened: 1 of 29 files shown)


NamelessCoder:
I think we may still need the escaping disabled on the loop and for ViewHelpers. It is very likely people will use HTML inside those loops and we don't want that escaped, so that needs to be confirmed.

monosize:
ok, the parent class AbstractLoopViewhelper has now escaping disabled.

Classes/ViewHelpers/Iterator/ImplodeViewHelper.php
```diff
@@ -26,11 +26,6 @@ class ImplodeViewHelper extends ExplodeViewHelper implements CompilableInterface
     protected static $method = 'implode';
 
     /**
-     * @var boolean
-     */
-    protected $escapeOutput = false;
```
NamelessCoder: We should skip this one; v:iterator.implode may be used to implode HTML contained in an array. Come to think of it, the right thing is probably to set escaping off in ExplodeViewHelper (which will also apply to ImplodeViewHelper).

monosize: I removed this one, because it's still in parent class ExplodeViewHelper defined.
https://github.com/FluidTYPO3/vhs/pull/1226/commits/a01c5f79feb0e651186a7b94cb26a502ad1fc795#diff-d53e03043a80b21e7dd7bce2838ba895R35

PR #1229, opened 2017-02-08
[BUGFIX] FormatViewHelpers - TYPO3v8+ escaping

monosize:


NamelessCoder:
@monosize sorry it took a while to get around to this one - I think we need to set `$escapeChildren = true` (or just not override that one) for pretty much all of these ViewHelpers. If we don't, any variable that gets passed from a potentially unsafe source will be explicitly not escaped, with no way for the template editor to control that. In this case I think it's best if we preserve how Fluid would normally render variables and then set `$escapeOutput = false` to prevent potential double escaping. That way, it still has to be a conscious decision to nest/wrap `f:format.raw` if you wish to allow HTML in a specific variable you pass to any of these ViewHelpers.

Or am I missing a detail about the use case or previous behavior?

NamelessCoder:
Closing due to inactivity, but feel free to reopen it if it becomes relevant again ;)

PR #1232, opened 2017-02-08
[BUGFIX] Set escaping behavior in Render ViewHelpers

### LINKED ISSUE #1228: Missings "arguments" in RequestViewHelper
Hi,

In the file "Classes/ViewHelpers/Render/RequestViewHelper.php", the argument "arguments" is missing.

`  $this->registerArgument('arguments', 'array', '...');`

monosize:
Added missing argument "arguments" for v:render.request Viewhelper - resolves #1228


NamelessCoder:
Needs a bit of conflict resolution - another commit in the meantime added the "arguments" argument on v:render.request so your patch should avoid adding that.

monosize:
Ok, changes on v:render.request have been reset.

PR #1234, opened 2017-02-08
[BUGFIX] Set escaping behavior in Page ViewHelpers

monosize:


NamelessCoder:
All of these should definitely escape HTML, so we don't want this particular change.

Rule of thumb: only when a VH produces HTML or a completely trusted value (for example, if we casted the return type to integer, float, boolean and so on) should we disable escaping. When a VH outputs things like strings from TS they should by default escape output (since we don't necessarily know that the value is safe).

monosize:
Ok, closing the PR.

PR #1239, opened 2017-02-08
[BUGFIX] Set escaping behavior in Asset ViewHelpers

monosize:


NamelessCoder:
Asset VHs don't produce any output and the escaping that $escapeOutput controls is never applied when those assets get rendered. Will update this to $escapeChildren which seems more appropriate.

PR #1249, opened 2017-02-10
[BUGFIX] respects absRefPrefix

torvitas:


torvitas:
See https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/extbase/Classes/Service/ImageService.php#L91

Classes/ViewHelpers/Media/SourceViewHelper.php
```diff
@@ -165,6 +178,8 @@ public function preprocessSourceUri($src)
         } elseif ('BE' === TYPO3_MODE || false === (boolean) $this->arguments['relative']) {
             if (GeneralUtility::isValidUrl($src)) {
                 $src = ltrim($src, '/');
+            } elseif ($this->environmentService->isEnvironmentInFrontendMode()) {
```
NamelessCoder: `TYPO3_MODE === 'FE'` would do the trick. If we can avoid an injected dependency, we should do so ;)

torvitas: I tried to use the api over a global constant like they do at https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/extbase/Classes/Service/ImageService.php#L91

Going to change it to use the constant.

torvitas:
Hi, is there anything else to do before merge?

The travis checks that failed do not look like they are related to this pr.
"Fatal error: Class 'TYPO3\CMS\Core\Tests\UnitTestCase' not found in /home/travis/build/FluidTYPO3/vhs/Tests/Unit/AssetTest.php on line 22"

NamelessCoder:
Sorry about the failing tests - they were indeed not caused by you!

PR #1251, opened 2017-02-13
[BUGFIX] Consider backend context

### LINKED ISSUE #1250: v:page.info crashes in TYPO3 backend
When using the `<v:page.info>` view helper in the fluidcontent preview section, it crashes:

> Call to a member function getPage_noCheck() on null

It tries to access `$GLOBALS['TSFE']` which does not exist in the backend:

     $page = $GLOBALS['TSFE']->sys_page->getPage_noCheck($pageUid);

bjo3rnf:
Fixes #1250

### DIFF
--- a/Classes/ViewHelpers/Page/InfoViewHelper.php
+++ b/Classes/ViewHelpers/Page/InfoViewHelper.php
@@ -9,7 +9,9 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Page;
  */
 
 use FluidTYPO3\Vhs\Traits\TemplateVariableViewHelperTrait;
+use FluidTYPO3\Vhs\Utility\ErrorUtility;
 use TYPO3\CMS\Fluid\Core\ViewHelper\Facets\CompilableInterface;
+use TYPO3\CMS\Frontend\Page\PageRepository;
 use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
 use TYPO3\CMS\Fluid\Core\Rendering\RenderingContextInterface;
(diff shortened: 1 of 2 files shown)


bjo3rnf:
I have no idea especially because of the static context.

bjo3rnf:
@NamelessCoder can we simply ban using this vh in the backend by law or something?

NamelessCoder:
It's times like these when I wish that https://review.typo3.org/#/c/48580/ had become a reality...

Can we ban use of this in the backend? Yes, we can do that. You can for example make the VH throw an exception if TSFE does not exist. Some day perhaps I'll lift the simulation stuff from the merge request above, over to VHS to give a generic way to simulate FE for *any* ViewHelper which needs it. But until that day, I'm okay with any VH that demands FE context throwing an exception if the FE dependencies it needs aren't there (tec note: it must not check `TYPO3_MODE` but must check each dependency such as presence of `$GLOBALS['TSFE']`).

In any case: an Exception is a lot better than a fatal error trying to call methods on an object that isn't there. I'd be more than okay with a PR that implements such Exception throwing!

PS: we have a few VH that do internal FE simulation, leave those as-is. Don't start implementing simulation elsewhere ;)

cedricziel:
I vote for an exception that burns the house and leaves a polite note "this wouldn't have happened if you didn't use me in BE".

NamelessCoder:
IMHO the exception message should be along the lines of `sprintf('ViewHelper %s does not work in backend context without a simulated frontend.', static::class);` - then I'll adjust that message to point to `v:be.simulateFrontend` or whatever it will be called, if/wen that becomes available.

bjo3rnf:
I made some adjustments. Hope you like them. They are great. Believe me. Enjoy. ^^

PR #1258, opened 2017-02-21
[BUGFIX] ensures page overlay is used if enabled, fixes #1257

### LINKED ISSUE #1257: Resource/Record/FalViewHelper does not respect PAGES_OVERLAY
This leads to the fal records that are assigned to the original record are being displayed instead of those assigned to the correct overlay.

torvitas:
fixes #1258

### DIFF
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -73,6 +73,8 @@ class FalViewHelper extends ResourcesFalViewHelper
             $localisation = $pageRepository->getPageOverlay($record, $this->getCurrentLanguageUid());
             if (is_array($localisation)) {
                 $record = $localisation;
+                $record['uid'] = $record['_PAGES_OVERLAY_UID'];
+                $this->arguments['table'] = 'pages_language_overlay';
             }
         }
         return $record;
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -73,8 +73,6 @@ class FalViewHelper extends ResourcesFalViewHelper
             $localisation = $pageRepository->getPageOverlay($record, $this->getCurrentLanguageUid());
             if (is_array($localisation)) {
                 $record = $localisation;
-                $record['uid'] = $record['_PAGES_OVERLAY_UID'];
-                $this->arguments['table'] = 'pages_language_overlay';
             }
         }
         return $record;
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -118,7 +118,7 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
 
         $fileReferences = [];
         if (empty($GLOBALS['TSFE']->sys_page) === false) {
-            $fileReferences = $this->getFileReferences($table, $this->getField(), $sqlRecordUid);
+            $fileReferences = $GLOBALS['TSFE']->sys_page->getFileReferences($table, $this->getField(), $record);
         } else {
             if ($GLOBALS['BE_USER']->workspaceRec['uid']) {
                 $versionWhere = 'AND sys_file_reference.deleted=0 AND (sys_file_reference.t3ver_wsid=0 OR ' .
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -119,6 +119,10 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
         $fileReferences = [];
         if (empty($GLOBALS['TSFE']->sys_page) === false) {
             $fileReferences = $GLOBALS['TSFE']->sys_page->getFileReferences($table, $this->getField(), $record);
+
+            if ($fileReferences->count() == 0) {
+                $fileReferences = $this->fileRepository->findByRelation($table, $this->getField(), $sqlRecordUid);
+            }
         } else {
             if ($GLOBALS['BE_USER']->workspaceRec['uid']) {
                 $versionWhere = 'AND sys_file_reference.deleted=0 AND (sys_file_reference.t3ver_wsid=0 OR ' .
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -100,13 +100,12 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
      * getResources
      *
      * @param array $record
-     * @param string $uidFieldName
-     * @param string $table
      *
      * @return array
      */
-    public function getResources($record, $uidFieldName = 'l18n_parent', $table = null)
+    public function getResources($record)
     {
+        $uidFieldName = 'l18n_parent';
         $databaseConnection = $this->getDatabaseConnection();
         $table = $this->getTable();
         if (isset($record['t3ver_oid']) && (integer) $record['t3ver_oid'] !== 0) {
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -105,7 +105,7 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
      *
      * @return array
      */
-    public function getResources($record, $uidFieldName = 't3ver_oid', $table = null)
+    public function getResources($record, $uidFieldName = 'l18n_parent', $table = null)
     {
         $databaseConnection = $this->getDatabaseConnection();
         $table = $this->getTable();
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -105,7 +105,6 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
      */
     public function getResources($record)
     {
-        $uidFieldName = 'l18n_parent';
         $databaseConnection = $this->getDatabaseConnection();
         $table = $this->getTable();
         if (isset($record['t3ver_oid']) && (integer) $record['t3ver_oid'] !== 0) {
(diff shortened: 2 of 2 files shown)


Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -103,15 +103,19 @@ protected function getFileReferences($table, $field, $uid)
     public function getResources($record)
     {
         $databaseConnection = $this->getDatabaseConnection();
+        $table = $this->getTable();
         if (isset($record['t3ver_oid']) && (integer) $record['t3ver_oid'] !== 0) {
             $sqlRecordUid = $record['t3ver_oid'];
+        } elseif (isset($record['_PAGES_OVERLAY']) && true === $record['_PAGES_OVERLAY']) {
```
NamelessCoder: This is a pages-table specific operation and doesn't belong in the generic FAL ViewHelper. If anywhere, it should go into `v:page.resources.fal` ViewHelper. And this ViewHelper already does language overlay (see https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Page/Resources/FalViewHelper.php#L73) so I'm not sure if your problem is caused by using the incorrect ViewHelper (using generic one when you need the pages one)?

pixelplant: The ViewHelper already does language overlay and it successfully overlays data, but fails to retrieve the correct image from sys_file_reference since it still points to table "pages" in its arguments, while it should switch it to "pages_language_overlay" while on a translated page.
You can try it out, especially with slides enabled, just do a simple 
```
<v:page.resources.fal as="images" limit="2" slide="-1">
<f:image src="{images.0.uid}" treatIdAsReference="1" width="640c" height="470c" />
</v:page.resources.fal>
```
on your page, and use a different header image for the default language and a translated one.

So... when you fetch the records https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Page/Resources/FalViewHelper.php#L102 it looks for corresponding records in sys_file_reference but has the value  "pages" in $this->getTable() and not "pages_language_overlay" as it should have.

I guess a simple fix would be to add this line, right after the if: https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Page/Resources/FalViewHelper.php#L72
```
$this->arguments['table'] = 'pages_language_overlay';
```
but then we still need to provide the correct record id, as it uses the uid of the media defined in the parent page, and not the one from the overlayed one.

torvitas: @NamelessCoder I see your point that this must not be implemented in the generic FAL ViewHelper. We are actually using the Page/Resources/FalViewHelper but as this one just calls to parent::getResources in getSlideRecordsFromPage we tried to fix it there.

Our proposed solution would be:
- change the signature of Resource/Record/FalViewHelper::getResources to
```
public function getResources($record, $uidFieldName = 't3ver_oid', $table = null)
```
follow by somthing like that:
```
        $databaseConnection = $this->getDatabaseConnection();
        if (null === $table) {
            $table = $this->getTable();
        }                                                                                                          
        if (isset($record[$uidFieldName]) && (integer) $record[$uidFieldName] !== 0) { 
            $sqlRecordUid = $record[$uidFieldName];
        } else {
            $sqlRecordUid = $record[$this->idField];
        }  
```
- change Page/Resources/FalViewHelper::getSlideRecordsFromPage to call parent::getResources with the appropriate arguments if we need resources for an overlay

torvitas:
Hi, is there anything else I can do to improve this PR?

Classes/ViewHelpers/Page/Resources/FalViewHelper.php
```diff
@@ -80,10 +80,13 @@ public function getRecord($id)
...
     /**
      * @param array $record
+     * @param string $uidFieldName
+     * @param string $table
+     *
      * @return array
      * @throws \Exception
      */
-    public function getResources($record)
+    public function getResources($record, $uidFieldName = 't3ver_oid', $table = null)
```
torvitas: I just know it isn't the correct field if you have an page overlay.

TheKvist:
Hi there, just wanted to bump a bit and ask if there is anything we can do to get this PR through? We really need this fix in our installation and this does the job.

If you need any further changes to the code, I'd be glad to help out, I just need to know what has to be done.

@NamelessCoder With "l10n original field", do you mean [this](https://docs.typo3.org/typo3cms/extensions/core/8-dev/singlehtml/Index.html#feature-78169-introduce-translation-source-field-for-tt-content)? If so, I could add a check if this field is set for the TCA of the table in  question and get the correct result based on the configuration.
Is using t3ver_oid recommended as a fallback and if not, what field would you think is more appropriate?

NamelessCoder:
@TheKvist yep that's the one I mean. And the suggested solution sounds good; check + fallback. But please check that it is in fact `t3ver_oid` because as far as I know, this field is (also) used to store the original record UID when creating workspace versions regardless of language.

TheKvist:
@NamelessCoder I've added the proposed changes, do  you think this is okay so far?

NamelessCoder:
Sorry, am tied up with other tasks. Regarding the patch the one thing I'm not too happy about is the function signature change - if we merge that, the next release has to be a major release and a breaking change, so it may end up sitting here for a while (unless you can figure out a solution that doesn't change those signatures).

TheKvist:
I'm sorry, I honestly wasn't aware of that. I changed the signature back to its original form. Since we still check for the TCA setting regarding the translation parent of the record and l18n_parent is the field used by convention we can safely move it into the method scope.
The table can also be safely moved because the ViewHelper will throw an Exception if the table argument is not supplied. Thus we can retrieve it by getTable without problems.

TheKvist:
I hate to annoy, but any news on this one yet?

NamelessCoder:
@TheKvist One method still has signature changes and inside that method, an inherited method is called with two undeclared arguments which probably need to be handled differently, which is what I gathered you wanted to do in `getTable`.

This gave me the impression you were still working on it. Hence, no news on this one except what you yourself have to tell ;)

TheKvist:
Finally came back to this today.

I found that when a record has a localization and we set its uid to the value of its '_PAGES_OVERLAY_UID', then setting the table argument to 'pages_language_overlay' as suggested by @pixelplant, we can safely call `getSlideRecordsFromPage`, since the resources it returns are directly out of the FileRepository's `findByRelation` method, which - according to the doc comments - explicitly asks for the localized uid of the record.
So if I'm not mistaken, passing the '_PAGES_OVERLAY_UID' value as uid is the way to go if the FileRepository should actually fetch localized resources for pages.

TheKvist:
Don't merge yet, I've encountered the issue that after the table argument gets overridden in the view helper it's not possible to reliably reset it yet, resulting in the page repository trying to fetch more fields from pages_language_overlay than initially intended which breaks a lot.

NamelessCoder:
General tips:

* Try not to mutate the state of the VH itself
* Try not to mutate records (overwriting UID in particular) but instead use conditions to select the proper column
* Definitely don't mutate the VH arguments

If you avoid these you will be much less likely to run into state issues. Remember that the same VH instance may be called multiple times with different arguments (same VH instance is used if for example you have a loop with a VH inside) - and it is a lot easier to migrate the VH to static calling and improved performance later, if the state does not get mutated.

TheKvist:
@NamelessCoder  I've just rebased and checked and as far as I can tell, there is still no overlay done. However, if you look at my most recent commits you will see that I've noticed in the Resource/Record/FalViewHelper, we first check for the existence of `$GLOBALS['TSFE']->sys_page` just to **not** use any of it, but instead use the FileRepository, which sys_page would've called by default, too.

All in all I think it wasn't the intention here to check for the existence of sys_page, then use a different object to perform its task.

So as a solution I propose to actually use the PageRepository we've checked for, which will essentially do the same as before **but** additionally do an overlay after getting the file references. I've tested this in our installation for a while not and could neither find nor think of any conflicts that may occur when using sys_page.

TheKvist:
@NamelessCoder I've looked at the failed travis build and have seen it fails due to deprecated method calls in unit tests with TYPO3_VERSION set to "dev-master", so I don't think this is related to my changes. I'd be glad if you could have a look at them once again so we can resolve this PR.

Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -113,7 +117,7 @@ public function getResources($record)
 
         $fileReferences = [];
         if (empty($GLOBALS['TSFE']->sys_page) === false) {
-            $fileReferences = $this->getFileReferences($this->getTable(), $this->getField(), $sqlRecordUid);
+            $fileReferences = $GLOBALS['TSFE']->sys_page->getFileReferences($table, $this->getField(), $record);
```
NamelessCoder: Just for the record: `+ GLOBALS['TSFE']->sys_page` is actually the only change this patch introduces at this point, we don't even need to change an entire line. The rest is, I'm sorry to say, just noise. When you contribute to open source projects you should always consciously try to reduce the unrelated changes.

It helps us because we don't have to review noise and can focus on the signal.

In this case, please remove the unrelated changes so the actual change becomes clear. This means, don't modify the PHPDoc method (and in particular, don't add redundant PHPdoc comments that just echo the function name or change whitespace!). And although I see the point of doing so, please avoid extracting `$this->getTable()` to a variable (or open a second PR when we're done with this change).

You can ignore the deprecation warnings in the tests for now. Those will of course be fixed before this patch is merged - they are not caused by your change.

TheKvist:
My apologies, the rebase must've gone wrong in some way or another as it was actually my plan to have this line as only change. I've removed the lines in question to their original status, so the one you reviewed is the only change left.

As for this one, not only `+ GLOBALS['TSFE']->sys_page` is changed, but the third parameter to the method call needs to be changed as well, as the method called requires the third parameter to be an array holding a parent record for file references to be fetched, whereas the ViewHelper requires it to be the record's uid

NamelessCoder:
Thanks @TheKvist - looks a lot cleaner now ;)

NamelessCoder:
Pretty sure this is now solved by https://github.com/FluidTYPO3/vhs/commit/4930466ff8a9f281d971cccb99ba35526984f8aa. Sorry for not merging this pull request which you spent a lot of time on - but at the end we had three or more pull requests aiming to solve the same issue and the one in the link appeared to be the more complete solution.

If something is solved by this patch that isn't solved by the other, don't hesitate to let me know and I'll reopen/adapt this patch!

PR #1263, opened 2017-03-01
ensures that typo3-ter dependencies are fulfilled even if you install

### LINKED ISSUE #1262: fluidtypo3/vhs does not replace typo3-ter/vhs
So if I try to install via git or packagist it also installs typo3-ter/vhs if there are any dependencies to it, other TYPO3 extensions have solved that by adding a replace property to their composer.json.

torvitas:
from github or packagist. fixes #1262


cedricziel:
Thank you for your contribution! :)

Issue #1270, opened 2017-03-14
v:iterator.sort does not work when cached


cweiske:
v:iterator.sort does not work at all when run on a cached page. It only works once when I emptied `typo3temp`.

When removing CompilableInterface it works on every run.

Tested on vhs 4.0.0 and 4.1.0.

----

 I guess it needs to be converted to `renderStatic()` and `CompileWithContentArgumentAndRenderStatic`.

PR #1271, opened 2017-03-14
[BUGFIX] Make iterator.sort work when cached

cweiske:
Resolves: #1270 

Please review this. It works here, but I most probably did not test everything.


NamelessCoder:
Trait is safe to use - passing arguments this way is exactly the preferred way to convert non-static to static. Thanks for taking care of it!

PR #1276, opened 2017-03-19
[BUGFIX] Avoid overwriting variable

### LINKED ISSUE #1217: L ViewHelper does not respect extensionName argument
It's simple matter of order of lines in 

vhs/Classes/ViewHelpers/LViewHelper.php:75 we have

```
  $arguments = $arguments['arguments'];
  $extensionName = $arguments['extensionName'];
```

Instead we should have

```
  $extensionName = $arguments['extensionName'];
  $arguments = $arguments['arguments'];
```

Or even better find a different name for $arguments to avoid confusion.

bjo3rnf:
Fixes #1217


NamelessCoder:
Thanks @bjo3rnf - that one survived way too long.

PR #1279, opened 2017-03-24
[BUGFIX] Early return on empty menu

### LINKED ISSUE #1278: Page/Menu/BrowseViewHelper does not properly handle empty menu data
In my page template I have this:

```
<v:page.menu.browse as="pages" doktypes="DEFAULT,LINK,SHORTCUT,MOUNTPOINT,111">
```

This crashes when parent of current page has only "not in menu" child pages:

structure

This may be fixed with a simple test within `\FluidTYPO3\Vhs\ViewHelpers\Menu\BrowseViewHelper::render`:

```
        $menuData = $this->getMenu($parentUid);
        if (empty($menuData)) return null;
```

since `$menuData` is null for such a use case.

bjo3rnf:
Fixes #1278

### DIFF
--- a/Classes/ViewHelpers/Menu/BrowseViewHelper.php
+++ b/Classes/ViewHelpers/Menu/BrowseViewHelper.php
@@ -90,6 +90,9 @@ class BrowseViewHelper extends AbstractMenuViewHelper
         $parentUid = (integer) (null !== $this->arguments['pageUid'] ? $pageUid : $currentPage['pid']);
         $parentPage = $this->pageService->getPage($parentUid, $showAccessProtected);
         $menuData = $this->getMenu($parentUid);
+        if (empty($menuData)) {
+            return !empty($this->arguments['as']) ? $this->renderChildren() : '';
+        }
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Menu/BrowseViewHelper.php
```diff
@@ -90,6 +90,9 @@ public function render()
         $parentUid = (integer) (null !== $this->arguments['pageUid'] ? $pageUid : $currentPage['pid']);
         $parentPage = $this->pageService->getPage($parentUid, $showAccessProtected);
         $menuData = $this->getMenu($parentUid);
+        if (empty($menuData)) {
```
NamelessCoder: We may need to return either an empty string or if the `as` argument is used, return `$this->renderChildren()` without assigning the variable. Otherwise, custom rendering of empty menus gets disabled (no option to render a "no menu items" message for example).

bjo3rnf: So true, thanks.

PR #1282, opened 2017-03-30
[BUGFIX] Fix uid problem using FileReferences

BenjaminBeck:
This fixes a bug where the uid of a FileReferece object has been treated as uid of a File object.


BenjaminBeck:
Don't merge yet - there might be a problem ..

BenjaminBeck:
This can be merged - the problem had nothing to do with this change.

NamelessCoder:
@BenjaminBeck could you describe what happens before/after this change and how to reproduce the problem?

Classes/ViewHelpers/Media/Image/AbstractImageInfoViewHelper.php
```diff
@@ -88,7 +88,7 @@ public function getInfo()
 
         if (is_object($src) && $src instanceof FileReference) {
             $src = $src->getUid();
-            $treatIdAsUid = true;
+            $treatIdAsUid = false;
             $treatIdAsReference = true;
```
cedricziel: We should return the info once we know it's a reference. This would make the remaining code less obscure. - What do you think?

NamelessCoder: I think we would probably just be duplicating the code from lines 96-102 below. It's easier (perhaps not more legible though...) to let the code continue to the block below so I'm fine with this as-is.

PR #1306, opened 2017-04-23
[BUGFIX] Insert placeholder before tag's last occurence

### LINKED ISSUE #1283: insert JS assets at end fails if there is are head or body-tags inside js inside content
We have a javascript which generates a popup-page for printing-purposes. The js looks something like this:
`
...
write('<!doctype html><html><head></head>');
write('<body onload="print();" style="font-family: Arial; font-size: 15px;">');
write(text);
write('</body></html>');
...
`
But sadly VHS corrupts this javascript by adding his <script type="text/javascript" src="typo3temp/vhs-assets-pageslidejs.js?1491303894"></script> before the closing body. Also it destroys the first line, because it adds a linebreak between <head> and </head>.

I had to remove the head-tags and I do not close the body-tag, so it works for me for now. But maybe you can rethink your "find the end of the document"-method-stuff.

Thanks.

bjo3rnf:
Fixes #1283

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -226,7 +226,9 @@ class AssetService implements SingletonInterface
         if (false === strpos($GLOBALS['TSFE']->content, $assetMarker)) {
             $inFooter = (boolean) (false !== strpos($markerName, 'Footer'));
             $tag = true === $inFooter ? '</body>' : '</head>';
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks @bjo3rnf - when you come around, would you mind removing me as contributor for your fork of the repo? I no longer need access since we can push to PR branches ;)

PR #1310, opened 2017-04-27
[FEATURE] Make column optional for content ViewHelpers

cvholten:
Get possibility to get all columns by settting argument "column" to "-1". Maybe there is a better solution. Sorting would not work anymore.


NamelessCoder:
Small note: the right way here would be to make the column argument optional and simply not include `colPos` in the condition if argument is null.

cvholten:
Maybe know it´s fine?

NamelessCoder:
Thanks for the addition - and welcome to the list of contributors! 🎉

PR #1313, opened 2017-04-28
[BUGFIX] Wrong $languageMenu entries when using "order" with non existing lang uids

kszymukowicz:
When "order" is used in v:page.languageMenu and one of the languages from "order" does not exist (for example because is disabled in backend) then its anyway added to $languageMenu array. That leads to problems for example with RealUrl 2.2.0 which disables frontend cache when there is a try to create a link with non existing language.


NamelessCoder:
Thanks for the patch and welcome to the list of official contributors ;)

kszymukowicz:
Thanks for merging! :)

PR #1316, opened 2017-05-11
[BUGFIX] Respect access config in asset service

tantegerda1:
Respect the config settings
- TYPO3_CONF_VARS / SYS / createGroup
- TYPO3_CONF_VARS / SYS / fileCreateMask
- TYPO3_CONF_VARS / SYS / folderCreateMask

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -619,6 +619,7 @@ class AssetService implements SingletonInterface
                 } else {
                     if (false === file_exists($temporaryFile)) {
                         copy($realPath, $temporaryFile);
+                        GeneralUtility::fixPermissions($temporaryFile);
                     }
                     $replacements[$matches[1][$matchCount]] = $wrap[0] . $temporaryFileName . $suffix . $wrap[1];
                 }
@@ -734,7 +735,7 @@ class AssetService implements SingletonInterface
      */
     protected function writeFile($file, $contents)
     {
-        file_put_contents($file, $contents);
+        GeneralUtility::writeFile($file, $contents);
     }
 
     /**


tantegerda1:
Resolves #1316

Classes/Service/AssetService.php
```diff
@@ -734,7 +735,7 @@ public function clearCacheCommand($parameters)
      */
     protected function writeFile($file, $contents)
     {
-        file_put_contents($file, $contents);
+        GeneralUtility::writeFile($file, $contents);
```
cedricziel: Shouldn't we use the third parameter then?

tantegerda1: I'm not sure about that: When a file is newly created, the correct permissions are applied anyways. Setting the third parameter would _readjust existing wrong-permissioned files_ - which can be achieved by clearing cache/typo3temp too.

cedricziel: Well... fput would then only use the current umask - as file_put_contents does, wouldnt it? To do what you implied, applying the configured permissions, we would need the third parameter. IMHO.

ping @NamelessCoder

tantegerda1:
To come back to this: I *did* test my changes in a real-life project - with the unmodified pull request applied, _new_ temporary asset files are created with proper group/rights. The third parameter to `writeFile` is not necessary for new files, it only affects _already-existing_ files by forcing their group ownership/access rights after the write operation.
So we could re-phrase the question: Do we want to forcibly repair "broken" already-existing temporary files on-access, possibly sacrificing some performance? I'd vote against, because
- for affected setups, fixing the rights once would suffice (chgrp, chmod or just remove those temporary files)
- seems like the current situation is not a problem for many hosting constellations (I could not find any prior relevant reports)

Many thanks for your thoughts and continued efforts towards vhs. Much appreciated!

cedricziel:
Let's go then :)

Thanks a bunch!

PR #1318, opened 2017-05-16
[BUGFIX] content.render fallback language

LudiYoo:
reference: fix language fallback #1135
> So the better approach would be to use the defined content fallback from config.sys_language_mode.
Condition: <v:content.render pageUid="{activePageUid}" column="4" />
Tested with langs (en, uk, fr): 
- en: default language with content ✓
- uk: not translated page => shows default content ✓
- fr: translated page => shows fr content ✓

### DIFF
--- a/Classes/ViewHelpers/Content/AbstractContentViewHelper.php
+++ b/Classes/ViewHelpers/Content/AbstractContentViewHelper.php
@@ -131,9 +131,15 @@ abstract class AbstractContentViewHelper extends AbstractViewHelper
             $order = $order . ' ' . $sortDirection;
         }
         $hideUntranslated = (boolean) $this->arguments['hideUntranslated'];
+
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Looks reasonable enough - @bjo3rnf / @cedricziel can you confirm this is the right way to implement fallback, e.g. that we don't need to perform checks on other configuration?

cedricziel:
I have added a commit that checks if the OL is set and then compiles the list of lang uids. Please double check.

LudiYoo:
Uncaught TYPO3 Exception
#1: PHP Warning: explode() expects parameter 2 to be string, array given
changes to implode()

Tested again with all languages; successfully

cedricziel:
Woops. Good catch. Though i could just inline edit and then got distracted.

cedricziel:

PR #1331, opened 2017-07-03
[BUGFIX] Allow correct serialization of UncacheTemplateView

cedricziel:
This is a moonshot to fix the serialization issue. It simply re-attaches the objectmanager and makes sure the renderingStack array gets serialized correctly so we can rely on it being an empty array and save the null check.


NamelessCoder:
It might not be the silver bullet that allows everything to be serialized on both 7.6 and 8.7+ but it is definitely a step in the right direction, to be more picky about what we do serialize and how we wake up the class.

Issue #1329, opened 2017-06-27
Reintroduce excludeSubpageTypes for v:page.menu.browse


SadlerJohnson:
We include a browse menu for each section of a 6000 page website. Around 250 of these pages are shortcuts for alternative spellings or synonyms. These shortcuts can redirect to the next article or an article in another branch of the page tree altogether. The issue is, that hitting these shortcuts via next and previous can take the user out of the current section or, worse, keep them in a loop within the section. The worst case scenario is when the synonym is directly after its respective main article.
I could exclude these 250 pages with excludePages but that would be cumbersome at best and requires editors to change the template when adding shortcuts. Therefore, I request that consideration is given to reintroducing excludeSubpageTypes which was deprecated after VHS 2.4.0.

NamelessCoder:
Not an unreasonable request given the use case. @bjo3rnf and @cedricziel what do you guys think? Shall we restore this or are we somehow prevented from that by the APIs we use?

@SadlerJohnson if you need a quick fix for this you could do a bit of manual rendering by pulling a standard menu and combining manual rendering with the `v:iterator.next`, `v:iterator.previous`, first, last and so on ViewHelpers to handle your next/prev linking. This is by the way a trick you can also use for pagination purposes - and I hope it helps in this case too.

bjo3rnf:
Agreed. Requires some refactoring though but I'll put it on the todo list.

PR #1335, opened 2017-07-15
[TASK] Re-implement excludeSubpageTypes argument

bjo3rnf:
Addresses #1329

### DIFF
--- a/Classes/Service/PageService.php
+++ b/Classes/Service/PageService.php
@@ -48,22 +48,28 @@ class PageService implements SingletonInterface
 
     /**
      * @param integer $pageUid
-     * @param array $excludePages
-     * @param boolean $includeNotInMenu
-     * @param boolean $includeMenuSeparator
-     * @param boolean $disableGroupAccessCheck
+     * @param array $options
      *
      * @return array
      */
-    public function getMenu(
-        $pageUid,
-        array $excludePages = [],
-        $includeNotInMenu = false,
-        $includeMenuSeparator = false,
-        $disableGroupAccessCheck = false
-    ) {
+    public function getMenu($pageUid, array $options = []) {
+        $excludePages = isset($options['excludePages']) ?
+            $options['excludePages'] : [];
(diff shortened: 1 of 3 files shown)


NamelessCoder:
I've had a bit of a think about this PR and https://github.com/FluidTYPO3/vhs/pull/1372 and I've come to the conclusion that the best way to handle this would be to pass the entire array of arguments to these functions, rather than extending the signature or creating an intermediate array to transfer arguments.

I've considered and opted against the following alternatives:

* Somehow making an `ArgumentCollection` that both defines and receives arguments for ViewHelpers, like a DTO that we can transfer. I chose not to since this very quickly over-complicated the VH structures.
* Inventing a "stupid" DTO rather than an array, with getters/setters to transfer things like requested menu configuration. I chose not to since this has very little difference to the $arguments array which we already have.

All in all I think a simple array is the most flexible method that isn't going to be too complex for contributors to understand. So my suggestion would be that in v5.0.0 we break these (internal) method signatures and switch to array as required type and then refactor as necessary to remove all other arguments than `array $arguments` from methods like `getMenu`.

And while doing so we should consider if this is the time to create services to return these things so we can begin cleaning up the class hierarchies and use simpler base classes!

bjo3rnf:
Yes, agreed. I had a discussion about that with @cedricziel already and he proposed http://symfony.com/doc/current/components/options_resolver.html which would be a quite elegant solution.

NamelessCoder:
That would work very well for our purposes, but I hesitate to introduce production composer dependencies until the time we pull the TER plug and go pure composer. Until then it's probably going to cause trouble for some users. Meanwhile we can at least implement safe fallbacks by using `array_replace($defaultsWeDefine, $arguments)` where this makes sense, for the methods we will refactor to receive arrays. Such defaults would make a nice static variable inside the method ;)

bjo3rnf:
I see and agree. I'll close this one for now and try to come up with something new soon.

NamelessCoder:
Thanks Björn, much appreciated :)

PR #1338, opened 2017-07-17
[BUGFIX] Multi language Support added for images in Frontend

cassini-mohsin:


NamelessCoder:
Thanks for the patch @cassini-mohsin - and welcome to the list of contributors! :tada

Issue #1330, opened 2017-06-27
[Q]<v:content.resources.fal return value


ogrosko:
I'm just wondering why VH `<v:content.resources.fal />` returns fal instance as `array` and not as `FileReference` object

I think it's much more easier this (assume image is instance of FileReference):
`{v:content.resources.fal(field: 'image') -> v:iterator.first() -> v:variable.set(name: 'image')}
  <f:image image="{image}"/>`

than this:
`{v:content.resources.fal(field:'image') -> v:iterator.first() -> v:variable.set(name: 'image')}
  <f:image treatIdAsReference="1" src="{image.id}" title="{image.title}" alt="{image.alternative}"/>`

I think it's this method [getResource](https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Resource/Record/FalViewHelper.php#L76) why it's simply does not return FileReference object as is?

NamelessCoder:
This ViewHelper is pretty old by now and this is from memory, so it may not be accurate, but as far as I remember this was done to give templates access to the actual record properties rather than only those properties existing on File or FileReference - to include fields you may have added to either table using TCA. Plus the ViewHelper is I think older than the capability on `f:image` to actually pass the File or FileReference object as input.

While it might not make sense to make a breaking change in this return value, perhaps it would be an easy matter to include the File or FileReference object as `{image.object}` using the vernacular of your example above. That would give you the object so you can pass that and avoid `f:image` loading it again, without changing the existing properties and breaking existing code. If you are up to it, a pull request to do this would be very welcome :)

ogrosko:
@NamelessCoder  Is this acceptable #1343 ?

ogrosko:
Check #1345 is it acceptable?

PR #1345, opened 2017-07-28
[TASK] - #1330 FalViewHelper resources as FileReference

ogrosko:
Resource Fal ViewHelpers returns FileReference instead of array of properties


NamelessCoder:
This is a very breaking change so we cannot accept it this way. While it might be that your particular use case calls for the actual FileReference object to be returned, every existing implementation is based on receiving an array.

At the very least we would need a toggle (argument) to control what gets returned, and the default behavior must be the current one.

PR #1350, opened 2017-08-05
[BUGFIX] Make substring viewhelper multibyte safe

### LINKED ISSUE #1348: v:format.substring: Problem with umlauts
I try to get the first character of strings with `{v:format.substring(content: '{string}', length: 1)}`. This works fine except for strings beginning with umlauts - the substring is empty. When I debug the result I get this: `'' (1 chars)` ...

I guess you could avoid this by using mb_substr instead of substr.

bjo3rnf:
Fixes #1348


cedricziel:
ext-mbstring ist often compiled as optional module. We should first check if the function is defined/mbstring is loaded instead of relying on it.

bjo3rnf:
Hm... mb_* is used in quite some viewhelpers without checks.

jdreesen:
Maybe require `symfony/polyfill-mbstring`?

cedricziel:
That would be the perfect solution, unfortunately composer-based solutions would only work for those instances using composer.


-- EDIT:

TYPO3 v8 already requires that polyfill, So I'm fine with the change.

NamelessCoder:
It appears that in 7.6.x the polyfill is only in `require-dev` so in order to preserve 7.6 compatibility I'd like to see a small (ternary) condition with `function_exists('mb_substr')` before assuming the function exists.

It would also be fine to implement `mb_` counterparts the same way in other places where we currently use non-mb-safe methods.

And once we bump to 8.7+ minimum requirement we can safely remove those checks.

cedricziel:
Well.. the mbstring polyfill is two files long. - What do you think about vendoring it into the extension for the time being? Would save some hassle and we could easily remove it on bumping the core requirement.

NamelessCoder:
> What do you think about vendoring it into the extension for the time being?

I hesitate to do that due to needing autoloaders, but if there's a clean solution that doesn't require new steps when creating TER releases, works in non-composer mode as well and doesn't conflict if the polyfill is already provided by TYPO3, then I'd be happy to consider it. Simply copying the polyfill files to the VHS repository could also be an option, though we'd probably have to load it through ext_localconf.php after detecting if the method is there.

cedricziel:
We would simply require the polyfill in ext_localconf from a local file if the function, let's say `mb_substr` doesn't exist. This wouldn`t collide with 8.7`s dependency, since this one is classmapped and included one step earlier with the composer autoloader.

NamelessCoder:
Sounds like a plan, assuming we just include those two files in the repository so there's no added requirements for creating releases.

cedricziel:
@NamelessCoder I proposed a change in https://github.com/bjo3rnf/vhs/pull/11 , please have a look there.

NamelessCoder:
Any special reason to check for two potential `mb_` methods before loading the polyfill (how can this use case happen - are there possibly other, partial polyfills)?

cedricziel:
It's taken from the original `symfony/polyfill-mbstring` as is - i presume they had a reason to put it into an additional condition.

cedricziel:
@NamelessCoder composer has such thing - it's a `files` directive that directly includes files - perfect for shimming - but it's more the missing composer support in TYPO3 and less a problem with composer.

Since TYPO3 core has limited support for autoloading, maybe we should add a `files` directive to the core and dump it with the classloader (https://github.com/TYPO3/TYPO3.CMS/blob/eb7539b6897e1a0b1199ebf79fe1804539d8c013/typo3/sysext/core/Classes/Core/ClassLoadingInformationGenerator.php#L211) ?

NamelessCoder:
Will "promote" this one to BUGFIX so it ends up in the change log - since everything is now UTF-8 by default in TYPO3, it's technically a bug if we use non-MB methods.

NamelessCoder:
Thanks for the fix guys - door is now open for PRs switching non-MB to MB methods throughout the repo ;)

PR #1351, opened 2017-08-09
[BUGFIX] Fix the build

cedricziel:
The PR tries to iteratively find a new way to build the project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 vendor
 .idea
 composer.lock
-/index.php
-/typo3
-/typo3_src
+/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,13 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+before_script:
+  - mkdir -p build/logs
+  - export TYPO3_PATH_WEB=$PWD/build/web
+
 install:
   - composer require --no-update "typo3/cms:$TYPO3_VERSION"
   - composer install --prefer-dist
-
-before_script: mkdir -p build/logs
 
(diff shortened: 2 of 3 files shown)


cedricziel:
Bingo, bongo.

NamelessCoder:
Guys, this is not so awesome. We don't BUGFIX the CI related parts, and it wasn't broken to begin with. I'd like it if you don't proceed with additional build changes until you've discussed them with me ;)

cedricziel:
It was broken indeed - additionally this changeset scopes the fallout of the typo3 cms composer scripts to a subdirectory.

What's wrong with the change itself, so I can maneuver around it the next time / fix it to your liking, apart from the tag of the commit message? :)

NamelessCoder:
The build was broken by a core change at some point but was restored about a week ago and has worked since - https://travis-ci.org/FluidTYPO3/vhs/builds. I understand the desire to not build this in the project directory, but the more special config and settings we add (especially the WEB path) the more we must expect to further change things later on, and the more stuff runs through typo3/cms-composer-installers which is undergoing some significant changes at the moment (not all of which I completely trust). They are as far as I am aware changing pretty much everything down to the naming of the settings.

The benefit of putting things in sub-dirs is imho outweighed by the need for this much additional settings/init combined with future expectations for the installer plugins - that's my main point of objection. Change's formals except for the commit message are fine, as usual, so it's the purpose rather than the shape of it ;)

cedricziel:
Heh. I for sure wasn't aware the builds turned green a while back, my bad. :)

I completely understand and share your thoughts re changes in the core scripts, but the customization is done the way the scripts expect it for a long time now. - I don't expect drastical changes except for v9, for which we would have to change things anyways.

Anyways - I'll wait for your feedback the next changes.

Issue #1359, opened 2017-08-28
page.languageMenu leads to SQL error


merzilla:
Please have a look into
https://github.com/FluidTYPO3/vhs/commit/c6767ab7e937d173ac0c518d75cdbe24d570d5d8#diff-d691c1059fef010d32dcbb2f4c243c72
there is an `if empty` instead of `if not empty` condition.

PR #1361, opened 2017-08-28
[BUGFIX] Fix if condition to check if array is *not* empty

merzilla:
Resolves: #1359


NamelessCoder:
Thanks - and congratulations on your first ever VHS patch! 🎉

Issue #1327, opened 2017-06-23
v:content.render


jwittorf:
Using:

* TYPO3 8.7.2
* VHS dev-development (https://github.com/FluidTYPO3/vhs/commit/40ea0d74bd54dc501f16bb3d5f5074127ef23f87)

When using `<v:content.render column="0"/>` this error occurs:

```
Argument 1 passed to FluidTYPO3\Vhs\ViewHelpers\Content\AbstractContentViewHelper::getRenderedRecords()
must be of the type array, null given, called in
L:\htdocs\t3.domain\typo3conf\ext\vhs\Classes\ViewHelpers\Content\AbstractContentViewHelper.php on line 111
```

It should be possible to just render the content of a specific column on the users current page. I've been using the simple form of `<v:content.render column="X" />` for quite some time now working like a charm and need it working again as it should.

The [docs](https://fluidtypo3.org/viewhelpers/vhs/development/Content/RenderViewHelper.html) don't indicate that anything has changed on how to use it except for the new possibility to `Render a single content element by its UID` but this is not what I'm intending to do and what the function should be able to do.

DanielSiepmann:
I do not get an error, but the whole content of a page when using column "0" in Version 4.3.0.
The reason is that there is a type cast and 0 is a number which in the used condition is false with fallback to render whole page.

NamelessCoder:
Hi guys - please try the fix in the linked pull request and report back if that prevents both of your issues! Thanks :)

dahaupt:
@NamelessCoder Hi there, I've got a different error message coming with version **4.3.0** using `<v:content.render column="0"/>`:

`#1257246929: Tried resolving a template file for controller action "Content->render" in format ".html", but none of the paths contained the expected template file ().`

With #1362 it's working again!

NamelessCoder:
@dahaupt The error you got there indicates you may have a content element that's incorrectly configured and only got rendered because of the bug with colPos 0. For example it could be in a colPos that isn't part of the layout. The patch in #1362 just hides the issue with that element by avoiding rendering it so be careful you don't trip over this other element in another context!

dahaupt:
@NamelessCoder Thanks for that information. I anyway had a mistake in my template using the `colPos` value for `column` in `<v:content.render column="0"/>`. I changed it using the name of the specific column and now it's working fine (reading docs carefully works 😄) . 

Sorry for the confusion in this matter.

NamelessCoder:
[...]

PR #1362, opened 2017-08-28
[BUGFIX] Add additional sanity checks for colPos and uid-lists

NamelessCoder:
Close: #1327

### DIFF
--- a/Classes/ViewHelpers/Content/AbstractContentViewHelper.php
+++ b/Classes/ViewHelpers/Content/AbstractContentViewHelper.php
@@ -150,11 +150,11 @@ abstract class AbstractContentViewHelper extends AbstractViewHelper
         $languageCondition .= ')';
 
         $contentUids = $this->arguments['contentUids'];
-        if (true === is_array($contentUids)) {
+        if (true === is_array($contentUids) && !empty($contentUids)) {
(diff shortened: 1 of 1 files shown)


bjo3rnf:
I can confirm this patch fixes the issue of all content elements with colPos 0 being unintentionally rendered.

cedricziel:
@NamelessCoder care to check what's wrong here? Seems the ViewHelper is lacking the property `viewHelperNode` and thus the tests can't iterate on the childNodes

NamelessCoder:
@cedricziel The composer deps are loose enough that we now get Fluid 2.4.0 so my guess would be that the VH is using some very old API that wasn't intended. Specifically it probably overrides a method it shouldn't or an additional method must be called from the unit test when building mock. If you have time, feel free to take a look - otherwise I'll take a look at the tests tonight.

cedricziel:
Thx for the fix itself, and the upstream fixes as well, @NamelessCoder !

NamelessCoder:
Was nothing, I wish I had caught this one earlier. Now the second problem with localisation config needs a solution so if you or someone else has time for that, please do take a look!

TheNaderio:
I have flux, fluidpages and vhs all in latest development branch - and a fresh TYPO3 8.7.6
but my contentelement still renders twice.
It makes no difference if it is in colpos 0 or 1

Issue #1364, opened 2017-08-30
htmlencoded src in Media/ImageViewHelper 


mikevinmike:
In version 4.3 FluidTYPO3\Vhs\ViewHelpers\Media\ImageViewHelper generates htmlencoded src with 
`<v:media.image src="fileadmin/Roadtrip-LP37-two-girls-peeking-out-car.jpg" alt="AAA" class="imgWidth100HeightAuto" width="1140c" height="814c" quality="200" title=""/>`

The outcome is 
`<img class="imgWidth100HeightAuto" width="1140" height="814" src="/fileadmin%2F_processed_%2F2%2F3%2Fcsm_Roadtrip-LP37-two-girls-peeking-out-car_68cb015f58.jpg" title="">`

This does not occur with version 4.2.

othmar52:
the value of the `srcset` attribute is also affected by this bug

supercosh:
In /Classes/ViewHelpers/Media/AbstractMediaViewHelper.php rawurlencode() is used, which escapes slashes to %2f, therefore URI's are not working anymore. This has changed from 4.2 to 4.3 and is breaking all viewhelpers inheriting from AbstractMediaViewHelper.

NamelessCoder:
If you have the time, please confirm if https://github.com/FluidTYPO3/vhs/pull/1367 solves the issue on your setup. Thanks :)

cushingw:
We encountered the same issue with 4.3.1 and <v:uri.image> and #1367 resolved the issue for us.

NamelessCoder:
Solved in dev-development, TER release is pending.

PR #1367, opened 2017-09-03
[BUGFIX] Don't urlencode image urls

bjo3rnf:
Addresses #1364. To be honest I don't remember why url encoding was used in the first place here so chances are this breaks something else. Testing and feedback is highly appreciated.

### DIFF
--- a/Classes/ViewHelpers/Media/AbstractMediaViewHelper.php
+++ b/Classes/ViewHelpers/Media/AbstractMediaViewHelper.php
@@ -58,7 +58,7 @@ abstract class AbstractMediaViewHelper extends AbstractTagBasedViewHelper
      */
     public static function preprocessSourceUri($src, array $arguments)
     {
-        $src = $GLOBALS['TSFE']->absRefPrefix . rawurlencode($src);
+        $src = $GLOBALS['TSFE']->absRefPrefix . $src;
         if (false === empty($GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_vhs.']['settings.']['prependPath'])) {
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Looks like this was added as part of a much larger change back in 2015:

https://github.com/bjo3rnf/vhs/commit/4db4ed567f74e3fef25c452fe5d6d337dc79c615#diff-dd3c40c66dd2d2443cb3e8be9e0cf6d6R48

If anything is broken by this patch I would presume it would be related to responsive images.

NamelessCoder:
Approved but don't merge until verified.

NamelessCoder:
FYI, the TYPO3 core change log explicitly says to use raw url encoding as substitute for that removed/deprecated method so if the problem is that those just behave differently (removed method didn't escape this exact byte) then I'd like to know so I can adapt the TYPO3 core change log message and explain the difference ;)

bjo3rnf:
> the TYPO3 core change log explicitly says to use raw url encoding as substitute for that removed/deprecated method

Ok, but I can't think of a good reason to encode a file's url in the first place.

bjo3rnf:
To quote myself:

> Ok, but I can't think of a good reason to encode a file's url in the first place.

I think I found one. The problem here was that the full path to the image was urlencoded which includes all slashes. There might be rare cases where the filename contains an already escaped slash which should be encoded of course. Should we consider this edge case as valid or leave as is @NamelessCoder?

NamelessCoder:
@bjo3rnf You answered the question with a pull request - best way to answer! :)

PR #1368, opened 2017-09-04
[BUGFIX] Correctly instantiate ObjectManager

### LINKED ISSUE #1369: v:render.uncache : Cannot instantiate interface TYPO3\CMS\Extbase\Object\ObjectManagerInterface
Hi,
I tried to use the v:render.uncache ViewHelper (TYPO3 8LTS, VHS 4.3) but I get the error:

Cannot instantiate interface TYPO3\CMS\Extbase\Object\ObjectManagerInterface

Error thrown in file
/Users/stephangrass/Documents/_Webs_/hs-mannheim.de/typo3/typo3_src-8.7.4/typo3/sysext/core/Classes/Utility/GeneralUtility.php in line 3893

cedricziel:
`ObjectManagerInterface` doesnt seem to be resolved correctly on wakeup. - As a workaround, and since ObjectManager is a singleton anyways and thus is resolved earlier, we can grab the prepared instance without using the interface here.

Ref: https://github.com/FluidTYPO3/vhs/commit/94c3d77daffa0f9ad65c3b6136f26f0131bab90f#commitcomment-24065966

Closes #1369

### DIFF
--- a/Classes/View/UncacheTemplateView.php
+++ b/Classes/View/UncacheTemplateView.php
@@ -11,6 +11,7 @@ namespace FluidTYPO3\Vhs\View;
 use TYPO3\CMS\Core\Cache\CacheManager;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Mvc\Controller\ControllerContext;
+use TYPO3\CMS\Extbase\Object\ObjectManager;
 use TYPO3\CMS\Extbase\Object\ObjectManagerInterface;
 use TYPO3\CMS\Fluid\Compatibility\TemplateParserBuilder;
 use TYPO3\CMS\Fluid\Core\Compiler\TemplateCompiler;
@@ -52,7 +53,7 @@ class UncacheTemplateView extends TemplateView
      */
     public function __wakeup()
     {
-        $this->objectManager = GeneralUtility::makeInstance(ObjectManagerInterface::class);
+        $this->objectManager = GeneralUtility::makeInstance(ObjectManager::class);
     }
 
     /**


stephangrass:
After trying the BUGFIX 'Correctly instantiate ObjectManager' by cedricziel I get the error:

The Fluid template files "" could not be loaded.

NamelessCoder:
ObjectManager or ObjectManagerInterface? Join us next week as we traverse this old mine field again! :)

Seriously though, it really sucks we can't predictably use the interface and expect an implementation. Looks like the specific implementation is the only safe choice when it comes to ObjectManager at least.

NamelessCoder:
@stephangrass The error you get there appears to be unrelated and was simply hidden beneath this one. You can enable debug mode in TYPO3 to get more info but usually it's because of a problem in template paths definitions or not having path definitions at all.

tantegerda1:
The phpdoc of `GeneralUtility::makeInstance` says
> You may want to use \TYPO3\CMS\Extbase\Object\ObjectManager::get() if you
> want TYPO3 to take care about injecting dependencies of the class to be
> created. Therefore create an instance of ObjectManager via
> GeneralUtility::makeInstance() first and call its get() method to get
> the instance of a specific class.

Well, this mentions just Dependency Injection. But I'm not sure to which extent `GeneralUtility::makeInstance` does Interface resolving. I trained myself to used `ObjectManager::get` whenever something that smells like _"uuh, extbase involved"_ should be instanciated.
In my experience this makes the difference when resolving `ObjectManagerInterface`.

cedricziel:
The point is (the code itself is pretty straight forward): the 'container' if we call GeneralUtility::makeInstance by what it does there, only knows about Implementations when they are explicitly registered. That's the true minefield here. ObjectManagerInterface has no registered implementations in the config array under SYS->Objects (beware of the casing massacre), hence we need to grab ObjectManager by its implementation.

I would sacrifice my first born child for a proper IoC container with a defined lifecycle and a compilation phase.

PR #1374, opened 2017-09-06
[BUGFIX] Use shortcut target page for active/subpage check

mbrodala:
This ensures that shortcut menu items are marked as active if the visitor is currently located on the shortcut target page.


mbrodala:
This actually doesn't help for a shortcut pointing to the root page since it will be marked as active all the time.

Issue #1379, opened 2017-09-12
SQL error in viewhelper LanguageMenu


morinfa:
Tag 4.3.1.
In class LanguageMenuViewHelper, function parseLanguageMenu :
```
        if (empty($limitLanguages)) {
            $where = 'uid IN (' . implode(',', $limitLanguages) . ')';
        } else {
            $where = '1=1';
        }
```

If $limitLanguages array is empty, the following where clause is generated :
`uid IN ()`

NamelessCoder:
This is already solved in dev-development; `!empty()` is now used.

PR #1381, opened 2017-09-12
[BUGFIX] Filter empty array indexes

cedricziel:
Crossing them fingers...

Fixes: #1379


NamelessCoder:
I don't think this one fixes #1379 (at least not all the way through) but even so, filtering to remove empty values sounds good to me since that could catch some issues with dynamic language lists.

Issue #1383, opened 2017-09-13
[BUG] Registration of VHS namespaces


ogrosko:
Registration of VHS namespaces in `ext_localconf.php` is not correct now
`$GLOBALS['TYPO3_CONF_VARS']['SYS']['fluid']['namespaces']['v'] = ['FluidTYPO3\\Vhs\\ViewHelpers'];`

ViewHelpers are not extensibile now (see feature [78145](https://forge.typo3.org/issues/78415))

I think it should be registered as follow:
`$GLOBALS['TYPO3_CONF_VARS']['SYS']['fluid']['namespaces']['v'][] = 'FluidTYPO3\\Vhs\\ViewHelpers';`

ogrosko:
Ok now I found that problem was in order of extensions in `PackageStates.php`
How can I register custom VHS namespace?

Because now when vhs ext. is last in PackageStates it overwrite my config.

Maybe:
`$GLOBALS['TYPO3_CONF_VARS']['SYS']['fluid']['namespaces']['v'][100] = 'Vendor\\SomeExt\\ViewHelpers\\Vhs';` or some another ideas?

NamelessCoder:
Make your extension either depend on or suggest VHS - that fixes the loading order issue. Since order has an effect here it would not be good to make the VHS registration of namespace add to an existing array because that would (at least in your specific use case w/o dependency/suggest) cause the VHS namespace to be added on top and as such, your ViewHelpers would be ignored anyway.

Logically your extension should depend on VHS since it overrides parts of VHS. Your other option is to declare the namespace in templates, e.g. `{namespace v=My\Ext\ViewHelpers}` - this adds to the namespaces without overriding existing ones and will always be added last regardless of dependencies' order.

The CMS Fluid adapter's ViewHelperResolver does not sort the default namespaces arrays according to priority but you're very welcome to open a feature request to suggest that: https://forge.typo3.org/projects/typo3cms-core/issues

PR #1384, opened 2017-09-13
[Bugfix] 1383 namespace-registration

ogrosko:
[BUGFIX] - #1383 Namespace registration does not allow to register other extensible ViewHelpres


NamelessCoder:
See https://github.com/FluidTYPO3/vhs/issues/1383 for an explanation why this patch will not have the desired effect, and for information how to correct your problem / suggest priority sorting as a feature in TYPO3.

Sorry to close this so quickly but hopefully you understand that the source of the problem isn't the registration but rather the dependency structure you've ended up with and that won't be fixed by this patch ;)

Issue #1392, opened 2017-09-29
page.resources.fal does not use pages_languages_overlay


cweiske:
When using `<v:page.resources.fal>` on a translated page, the files of the original page record are used and not the ones of the translated record.

This did once work, see issue #953.

liayn:
This change is breaking for us.
We relied on that because we inherit media from the default page

PR #1393, opened 2017-09-29
[BUGFIX] Correctly fetch media for translated pages

cweiske:
Utilize the PageRepository to correctly fetch resources of translated pages
that are stored in pages_language_overlay.

Resolves: https://github.com/FluidTYPO3/vhs/issues/1392

### DIFF
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -17,6 +17,7 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Database\DatabaseConnection;
 use TYPO3\CMS\Core\Resource\Exception\ResourceDoesNotExistException;
 use TYPO3\CMS\Core\Versioning\VersionState;
+use TYPO3\CMS\Frontend\Page\PageRepository;
 
 /**
  * Resolve FAL relations and return file records.
@@ -56,6 +57,11 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
     protected $fileRepository;
 
     /**
+     * @var \TYPO3\CMS\Frontend\Page\PageRepository
+     */
+    protected $pageRepository;
+
+    /**
      * @var boolean
      */
     protected $escapeOutput = false;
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -85,14 +92,23 @@ public function getResource($fileReference)
...
-     * @param string $table name of the table to get the file reference for
-     * @param string $field name of the field referencing a file
-     * @param integer $uid uid of the related record
+     * @param string $table  name of the table to get the file reference for
+     * @param string $field  name of the field referencing a file
+     * @param array  $record Database row
      * @return array
      */
-    protected function getFileReferences($table, $field, $uid)
+    protected function getFileReferences($table, $field, $record)
```
NamelessCoder: Just to protect against breaking changes for those that may have subclassed this ViewHelper: please rename this argument to `$uidOrRecord` and check/convert that as needed. If a UID is passed you can call `$this->getRecord($uidOrRecord)` to fetch the record.

Side note: we don't align the argument names/descriptions, please separate those with a single space.

Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -87,12 +94,27 @@ public function getResource($fileReference)
...
+    protected function getFileReferences($table, $field, $uidOrRecord)
     {
-        $fileObjects = $this->fileRepository->findByRelation($table, $field, $uid);
+        if (is_array($uidOrRecord)) {
+            $record = $uidOrRecord;
+        } else {
+            $record = $this->getRecord($uidOrRecord);
+        }
+
+        if ($table == 'pages') {
```
xf-: compare ===

cweiske: done.

cweiske:
@xf- @NamelessCoder - I've adjusted the patch.

Classes/ViewHelpers/Resource/Record/FalViewHelper.php
cweiske: done

cweiske:
patch rebased.

Issue #1397, opened 2017-10-04
Eliminate not working properly in 4.3.2 with TYPO3  8.7.7


GhanshyamBhava:
@NamelessCoder I created a variable which stores the string. I need to remove all the special character from the string and return output plain lower case.

`<v:variable.set name="var" value="{v:format.eliminate(content: '@string with special-charecter. !', characters:'@,!,#,$,%,^,*,-,:,?,.,ß' whitespace: 1, nonAscii: 1)}" />
<v:variable.set name="id" value="{v:format.case(string:'{var}',case:'lower')}" />`

I debug variable {id} but still return content value as it is

`<f:debug>{id}</f:debug>`

tantegerda1:
btw. the `characters` argument is described as

> Characters to remove. Array or string, i.e. {0: 'a', 1: 'b', 2: 'c'} or 'abc' to remove all occurrences of a, b and c

thus you don't need the comma separation (doing so will add _the comma_ to the list of eliminated characters).

PR #1399, opened 2017-10-04
[BUGFIX] Correctly format.eliminate multibyte chars

tantegerda1:
str_split() fails to consider multi-byte characters, while this ViewHelper will most probably be used to remove such characters.

Resolves: #1397

### DIFF
--- a/Classes/ViewHelpers/Format/EliminateViewHelper.php
+++ b/Classes/ViewHelpers/Format/EliminateViewHelper.php
@@ -130,7 +130,7 @@ class EliminateViewHelper extends AbstractViewHelper
         if (true === is_array($characters)) {
             $subjects = $characters;
         } else {
-            $subjects = str_split($characters);
+            $subjects = preg_split('//u', $characters, null, PREG_SPLIT_NO_EMPTY);
         }
         foreach ($subjects as $subject) {
             if (true === $caseSensitive) {
--- a/Tests/Unit/ViewHelpers/Format/EliminateViewHelperTest.php
+++ b/Tests/Unit/ViewHelpers/Format/EliminateViewHelperTest.php
@@ -158,6 +158,17 @@ class EliminateViewHelperTest extends AbstractViewHelperTest
     /**
      * @test
      */
+    public function removesMultibyteCharactersAsString()
+    {
+        $arguments = $this->arguments;
+        $arguments['characters'] = 'æ本';
+        $result = $this->executeViewHelperUsingTagContent('aäæå本bc', $arguments);
+        $this->assertSame('aäåbc', $result);
+    }
+
+    /**
+     * @test
+     */
     public function removesCharactersAsArray()
     {
         $arguments = $this->arguments;


NamelessCoder:
Thanks for the patch Ludwig!

Would you mind adding a couple of test cases in https://github.com/FluidTYPO3/vhs/blob/development/Tests/Unit/ViewHelpers/Format/EliminateViewHelperTest.php#L184 to prove that multibyte chars are now handled correctly in subject and replacements? ;)

tantegerda1:
I'm having some difficulties here:

* https://fluidtypo3.org/documentation/contributing/contribution-guide.html#the-quick-and-dirty-version does neither yield a `vendor/bin/make` nor `vendor/bin/runtests` - even if explicitely using `composer install --dev`. When run with `vendor/bin/phpunit --configuration phpunit.xml.dist --bootstrap Tests/bootstrap.php` a lot of tests fail (probably because I'm running them wrongly). Therefore, the new tests validity is pure speculation.

* Inexperienced with writing tests I fail to answer myself some basic questions:
  * Amend existing tests vs writing new ones (I'd rather adapt _many_ test cases to incorporate multibyte characters)
  * should one test (e.g. the affected `removesCharactersAsString()`) have multiple `->assert…()` statements or is it better to write a separate test (even if it tests the same functionality)

NamelessCoder:
Sorry about the rather outdated contribution guide. We've tried to simplify it a lot since then, so now the following is true for all repositories:

* Do `composer install --dev` in the cloned repository itself
* Run `./vendor/bin/phpunit` without arguments to perform unit tests
* If you wish to make sure your tests/changes also work on other TYPO3 versions, simply `composer require typo3/cms:dev-master`, `composer require typo3/cms:^7.6` and so on, then run phpunit again to verify.

Regarding the general test practice, the test for v:format.eliminate isn't exactly perfect. This is where it gets a bit theoretical, but generally speaking, unit tests should be testing public units (which nearly all ViewHelper tests do because they use the `executeViewHelper` helper method). Writing tests for VHS is usually easy enough, worst case scenario is you have to mock/inject some dependencies - of which there are plenty examples in VHS tests. The rule of thumb is: if there are tests already, try to extend those. If there's no test and you wish to add one, take an existing test that looks relevant (for example, another test in the same namespace) as base and adapt.

That's the general stuff and the easy part.

Then about writing _awesome_ tests: I personally appreciate that intention a lot, and I agree that this particular change could use more tests (but the one you wrote is, imho, sufficient for the time being). Taking the EliminateViewHelper as example I'd choose to refactor that so it uses a `dataProvider` and has a single test method which takes the VH arguments and the expected output as the only two arguments, then executes the VH with the arguments and asserts that the output is as expected. Although I can't point to a case right now you can simply search the Tests folder for `@dataProvider` and check any of the existing implementations. In short: the dataProvider annotation points to a method that returns an array of arrays of arguments that then each get passed to the test method. This is ideal when you want to quickly add a large number of tests where just the input arguments and expectations differ, but everything in the test method is the same - which it almost always is with ViewHelpers.

I would definitely not require that level of unit tests for this patch (in fact I'll merge it right after this comment) but if you want to practice, there are a couple of fairly simple ViewHelpers that aren't yet covered with tests and creating a dataProvider based test for the simplest of those would be excellent practice because you have the `executeViewHelper` and can focus on writing the input arguments and expectations.

Btw I'm also happy to teach you more about unit testing but judging from your patches so far, you're more than capable of learning this quickly by studying the examples in, for example, VHS ;)

tantegerda1:
@NamelessCoder, thanks for your didactic approach, i.e. encouraging me to provide the test (involving more back-and-forth), rather than opting for a quick do-it-self solution. And big thanks for taking the time to write such a detailed explanation! I appreciate that kind of knowledge transfer very much.

PR #1404, opened 2017-10-09
[BUGFIX] Check position before inserting assets at marker

mkempe:
Hey there,

first of all, many thanks for VHS!

This pull request checks the position, respectively if a `head` or `body` tag exists, before inserting assets at marker.

We have websites with "services", which render small portions of HTML with no `head` or `body` tags via page types, which should not contain assets. Before VHS 4.3.0 this was possible. With VHS 4.3.0 the assets get always inserted at the top.

I don't know, if this should be handled by VHS or better by the integrator, e.g. disable assets at all by certain page types?

Thanks again!


NamelessCoder:
Nice first patch - and welcome to the list of contributors! 🎉

mkempe:
@NamelessCoder Thank you 😊 and thanks again for all your hard work! 🎉

PR #1410, opened 2017-10-23
[BUGFIX] Find available languages

### LINKED ISSUE #1391: LanguageMenuViewhelper $inactive is always true
The LangaugeMenuViewhelper is broken, because $inactive in row 323 is always 1.
I debugged a lot, but wasn't able to find why inactive is always "true".

Something with 
 $inactive = in_array($key, $languageUids) || (integer) $key === $this->defaultLangUid ? 0 : 1;

xf-:
Resolves  #1391
THX @RozbehSharahi works for me.


brlocky:
Great fix ! works for me

Issue #1413, opened 2017-10-25
v:format.replace should have default value for substring or support "constant:NAMEOFCONSTANT"


Atsyn:
Will be better if this ViewHelper start support constants.
But will be fine if you switch argument "substring" to not require and set default value like PHP_EOL.

NamelessCoder:
Possible to achieve with `v:const`:

`substring="{v:const(name: 'PHP_EOL')}"`

Supports all constants and works for any ViewHelper argument.

NamelessCoder:
Closing the issue - referring to `v:const` will be the preferred solution when you need access to PHP constants in output or as arguments for other ViewHelpers.

PR #1414, opened 2017-10-25
Support constants in v:format.replace

neufeind:
Resolves: #1413

### DIFF
--- a/Classes/ViewHelpers/Format/ReplaceViewHelper.php
+++ b/Classes/ViewHelpers/Format/ReplaceViewHelper.php
@@ -25,8 +25,8 @@ class ReplaceViewHelper extends AbstractViewHelper
     public function initializeArguments()
     {
         $this->registerArgument('content', 'string', 'Content in which to perform replacement');
-        $this->registerArgument('substring', 'string', 'Substring to replace', true);
-        $this->registerArgument('replacement', 'string', 'Replacement to insert', false, '');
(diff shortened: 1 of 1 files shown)


neufeind:
Solution follows the code from "Iterator / ExplodeViewHelper". But if we want to support constants in more places maybe we should extract that tool-functionality to a central place?

NamelessCoder:
> But if we want to support constants in more places maybe we should extract that tool-functionality to a central place?

This sounds appropriate, if nothing else then to avoid code duplication. It would be fine to put this in a Utility that we then call on the argument value before using it as glue/delimiter/substring/whatever.

FYI, in the case of this particular ViewHelper and probably some others, I would require that this resolving becomes opt-in via a new argument, for example `resolveConstants`, so this addition does not change any current behavior. The reason being that in the implode and explode ViewHelpers it is trivial to allow this as you almost always split or glue with a single character - but in the replace and other ViewHelpers there's much greater chance someone intentionally searched for `constant:LF` and did not expect that this gets resolved to `\n` before replacement happens.

NamelessCoder:
We might also consider suggesting https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/ConstViewHelper.php as a valid alternative way to get this support in *every* ViewHelper. Using this in a ViewHelper argument would allow that exact resolving of any constant, e.g. `subject="{v:const(name: 'PHP_EOL')}"`.

NamelessCoder:
Won't add/fix - please use `v:const` for all your constant extraction needs ;)

PR #1425, opened 2017-11-27
[BUGFIX] enforce fixPermissions when writing asset files

peterkraume:
Sometimes TYPO3 caches can't be deleted with `typo3cms cache:flush --force` because of wrong file permissions of the EXT:vhs assets.
With this patch, `GeneralUtility:fixPermissions()` will be enforced to make use of `$TYPO3_CONF_VARS['BE']['fileCreateMask']`.

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -744,7 +744,7 @@ class AssetService implements SingletonInterface
      */
     protected function writeFile($file, $contents)
     {
-        GeneralUtility::writeFile($file, $contents);
+        GeneralUtility::writeFile($file, $contents, true);
     }
 
     /**


cedricziel:
Thx for the change, @peterkraume !

peterkraume:
👍 
Any idea when the next release of EXT:vhs is scheduled?

cedricziel:
There's no schedule yet - @NamelessCoder can we have a new bugfix release, pleeeeeeaaaaaaaaaaase? :shamrock:

cedricziel:
Not today, unfortunately - but maybe we can have a markdown file in the project that contains the release documentation? I think that's interesting for all of us who may tag and as far as I understood the release of the artifact is automated and there's not sensitive information - ayy?

NamelessCoder:
Perhaps over the weekend then? I'd also like to discuss the current release process as well because it was created in 2012 or so and could use an update, maybe even some automation!

True, there's no sensitive information so it can be public docs. But if you don't mind, let's discuss first, maybe design a cleanup and *then* do and document it ;)

cedricziel:
Alrighty - let's do this. I'll ping ya :)

NamelessCoder:
Great! I should be available from tomorrow around lunchtime and I've got the entire weekend off :)

peterkraume:
Guys, can we have a new release please?

PR #1426, opened 2017-11-28
Fix the build

cedricziel:


cedricziel:
@NamelessCoder could you please proof-read? Thx.

NamelessCoder:
As soon as time allows!

cedricziel:
No rush ;)

Classes/ViewHelpers/Extension/Path/SiteRelativeViewHelper.php
```diff
@@ -30,6 +31,10 @@ class SiteRelativeViewHelper extends AbstractExtensionViewHelper
      */
     public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
     {
+        if (true === version_compare('9.0', TYPO3_branch, '<=')) {
```
NamelessCoder: Is this right? Looks like this means that on 9.0+ we'll then be calling the deprecated, possibly removed, method below? Should that not instead perhaps target 7.6 specifically and call the deprecated method only on that version?

peterkraume:
I think this patch blocks the Travis checks in my PR #1425.
Would be great to get this one sorted out soon!

cedricziel:
Should be fine now for the purpose of the PR itself, but may not serve @peterkraume well, since TYPO3 master did lose render methods in the meantime ^^

peterkraume:
So what do I need to do to get #1425 merged?

PR #1427, opened 2017-12-15
[TASK] Pagination: retrieve GET variables

hlabbassi:


cedricziel:
Can you provide a short explanation what it does and why it matters?

hlabbassi:
We have a problem with "paginate" widget , we remain on first page since we do not retrieve the pagination parameter passed in GET of the URL, this modification allows to keep the current behavior by adding variables GET from the URL

cedricziel:
I'm generally ok with it - can you provide a simple test?

Classes/ViewHelpers/Render/RequestViewHelper.php
```diff
@@ -94,8 +96,14 @@ public static function renderStatic(
         $request->setControllerName($controller);
         $request->setPluginName($pluginName);
         $request->setControllerExtensionName($extensionName);
-        if ($arguments !== null) {
-            $request->setArguments($arguments);
+
+        $extensionService = static::getObjectManager()->get(ExtensionService::class);
+        $pluginNamespace = $extensionService->getPluginNamespace($extensionName, $pluginName);
+        $requestArguments = GeneralUtility::_GP($pluginNamespace) ?: [];
```
NamelessCoder: I'm not sure this is a good idea. This has the effect of always adding all URL arguments and replacing those you pass as arguments, and as such, completely changes how the request-render ViewHelper works: it no longer makes a fresh request, it also transfers anything you might have in the URL.

At the *very* least this should be opt-in, like declaring whether to include existing arguments when you use f:link.action or others. The default behavior should be to *not* transfer the existing arguments.

romm:
Hi,

I've changed a bit the behavior, this should make more sense now.

I also believe the name/description of the last commit should be used for the squashed commit (maybe a bit modified, but the main idea is described in here).

Please note that the commit also fixes an old bug, where the variable `$arguments` was overridden and could not be used within the `catch` block, leading to `graceful` and `onError` view helper arguments being useless until now.

NamelessCoder:
I hesitate a LOT to merge this one - although I see the reason for it, I definitely don't like the idea of letting the VH read request arguments directly from GET or POST, when it already allows you to set arguments. In theory you could assign your `Request` instance and use `arguments="{request.arguments}"` as a way to pass arguments.

The bug fix for $arguments being overwritten is very much desired, though. Can you separate that out into another patch?

NamelessCoder:
@hlabbassi Sorry for closing this but the concerns about passing arbitrary request arguments outweighs the benefit. There is one version of this feature that I would pretty much immediately accept:

Adding `request` with type `RequestInterface` as argument on `v:render.request`.

This would allow you to prepare the Request and assign it to template variables which is a much safer approach. Inside the ViewHelper then, when `$arguments['request']` is passed, this instance should be used instead of creating a fresh one through ObjectManager.

PR #1430, opened 2017-12-18
[TASK] Depend on typo3/cms-core rather than typo3/cms for v9 compat 

cedricziel:
Let's see how that goes..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_script:
   - export TYPO3_PATH_WEB=$PWD/build/web
 
 install:
-  - composer require --no-update "typo3/cms:$TYPO3_VERSION"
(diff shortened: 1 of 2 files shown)


NamelessCoder:
@cedricziel I think this dep is also set in `fluidtypo3/development`, but perhaps we can safely remove it from there / set that to `typo3/cms-core` as well.

cedricziel:
Re-visiting this, one problem is that CMS-Core is only available 7.6 and up. So should we burn the bridge just yet?

cedricziel:
*8.7 and up, corrected it above. Extra comment for e-mailers.

bmack:
@cedricziel can you rebase?

cedricziel:
@bmack updated the branch

NamelessCoder:
Implemented in https://github.com/FluidTYPO3/vhs/pull/1457

PR #1431, opened 2017-12-19
[BUGFIX] Fixed Problem of Multilanguage Images not working if fetched…

MohsinQK:
… from pages

### DIFF
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -107,13 +107,21 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
             $sqlRecordUid = $record['t3ver_oid'];
         } elseif (isset($record['_LOCALIZED_UID'])) {
             $sqlRecordUid = $record['_LOCALIZED_UID'];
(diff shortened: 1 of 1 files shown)


NamelessCoder:
@cassini-mohsin Thanks for the patch - meanwhile, I merged another patch for the FAL relation fetching, but it has a tiny difference compared to your patch. The other one does not check `_PAGES_OVERLAY_UID` in the record, so if you found a case where this is required (it seems from your pull request you did!) could you please describe that use case in more detail?

If you are certain the patch is still required, please rebase it on dev-development and resolve the conflicts. You should then end up with a patch that simply adds one more case to the two existing if-then-else blocks.

cassini-mohsin:
What I would like to achieve is to get Translated images if added on pages.
so the use case is, I have added some banner images from the page properties to slide that banner on every page, the banner has some English Text on it for English language and I have another image of the banner for the German Language with some German text on it. but currently, it was only fetching the image from the Default Language because it did not take care of the UID of the page that is translated in pages_language_overlay table.

Hope you will understand the Use Case.

NamelessCoder:
This is no longer needed - multi-language behavior for associated FAL assets is fully supported.

PR #1435, opened 2018-01-14
[TASK] Write publicly accessible files to typo3temp/assets/

ducrot:
Since TYPO3 8.0 publicly accessible files should be written to typo3temp/assets/.

[See TYPO3 Core Task #72580](https://forge.typo3.org/issues/72580)


NamelessCoder:
Thanks @ducrot - I labeled it as BUGFIX though technically it's not a bug, but this way it pops up in the changelog. You were completely right to label it TASK, though! ;)

PR #1441, opened 2018-01-31
Move class into the default image tag

fjacobi:
By adding the class to the image tag it won't actually style the image, but just the picture tag.

By moving it to the default image tag, it would allow to use the class to actually style the img.

Example:
https://codepen.io/fjacobi/pen/vpqzQG

### DIFF
--- a/Classes/ViewHelpers/Media/PictureViewHelper.php
+++ b/Classes/ViewHelpers/Media/PictureViewHelper.php
@@ -96,6 +96,10 @@ class PictureViewHelper extends AbstractTagBasedViewHelper
         $defaultImage = new TagBuilder('img');
         $defaultImage->addAttribute('src', $defaultSource);
         $defaultImage->addAttribute('alt', $this->arguments['alt']);
+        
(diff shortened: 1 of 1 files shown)


cedricziel:
Please excuse my ignorance, but wouldnt we want to be able to style both?

bjo3rnf:
IMHO not, because `picture` is only a wrapper for semantical/functional reasons same as `main` or `nav`. Mark the 'IMHO'.

fjacobi:
I think the note of the HTML spec is the best explanation for this:

> The picture element is somewhat different from the similar-looking video and audio elements. While all of them contain source elements, the source element's src attribute has no meaning when the element is nested within a picture element, and the resource selection algorithm is different. Also, the picture element itself does not display anything; it merely provides a context for its contained img element that enables it to choose from multiple URLs.
https://html.spec.whatwg.org/multipage/embedded-content.html#the-picture-element

bjo3rnf:
Word.

PR #1446, opened 2018-02-16
[BUGFIX] Handle crop argument when adding srcset variants

### LINKED ISSUE #1445: v:media.image does not honor cropping for srcset
`v:media.image` does not honor the `crop` argument for `srcset` variants: `src` is cropped, `srcset` images are not. The cause seems to be within `SourceSetViewHelperTrait`. A simple fix would be to fetch `$this->arguments['crop']` in `addSourceSet` and pass it as an additional parameter to `getImgResource` where it can be added to `$setup`.

This is still no decent crop handling (support for crop variants, no support for cropping information coming from the image `FileReference` itself, but only from the `crop` argument), but at least does fix one weirdness.

luzat:
This should fix #1445 for crop arguments (but not for cropping information coming from a `FileReference`).

### DIFF
--- a/Classes/Traits/SourceSetViewHelperTrait.php
+++ b/Classes/Traits/SourceSetViewHelperTrait.php
@@ -46,7 +46,7 @@ trait SourceSetViewHelperTrait
         $srcsetVariants = [];
 
         foreach ($srcsets as $key => $width) {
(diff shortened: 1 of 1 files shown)


electricretina:
I thought this was going to be in the latest development branch, it wasn't. I just patched the file on my server and all it working as expected! I hope to see this in the next release. Great Job!

Issue #904, opened 2015-09-18
v:page.menu hasSubpages on localized menus


lluvia:
Hi there,
i think there's a bug on the v:page.menu.
for a menu like this:
- Item 1 (localized)
      - Subitem 1.1 (not localized)
      - Subitem 1.2 (not localized)

<v:page.menu entryLevel="0" includeSpacers="1" useShortcutData="TRUE" levels="3" classFirst="first" classLast="last" expandAll="1">

renders the localized Item 1 with hasSubpages, even when the exist only in the original language

PR #1449, opened 2018-02-19
[BUGFIX] Set hasSubpages correctly on localized menus

h69:
this will fix a similar issue as in the previous closed issue https://github.com/FluidTYPO3/vhs/issues/904

for a menu like this:
- Item 1 (localized)
  - Subitem 1.1 (not localized)
  - Subitem 1.2 (not localized)

... the Viewhelper renders the localized "Item 1" with hasSubpages=true in any/all languages
This fix extends the submenu check inside parseMenu method and makes a recursive call (on parseMenu method) to check if the submenu isn't empty after processing. Only if the processed/parsed menu isn't empty we can be sure that the current menu item has sub-pages ;)

### DIFF
--- a/Classes/ViewHelpers/Menu/AbstractMenuViewHelper.php
+++ b/Classes/ViewHelpers/Menu/AbstractMenuViewHelper.php
@@ -441,9 +441,6 @@ abstract class AbstractMenuViewHelper extends AbstractTagBasedViewHelper
         $total = count($pages);
         $processedPages = [];
         foreach ($pages as $index => $page) {
-            if (true === $this->pageService->hidePageForLanguageUid($page)) {
-                continue;
-            }
             $count++;
             $class = [];
             $originalPageUid = $page['uid'];
--- a/Classes/ViewHelpers/Menu/AbstractMenuViewHelper.php
+++ b/Classes/ViewHelpers/Menu/AbstractMenuViewHelper.php
@@ -475,19 +475,12 @@ abstract class AbstractMenuViewHelper extends AbstractTagBasedViewHelper
                 $pages[$index]['current'] = true;
                 $class[] = $this->arguments['classCurrent'];
             }
-            // simply check if current page uid can have sub-pages
-            $subMenuPages = $this->getMenu($originalPageUid);
-            if (0 < count($subMenuPages)) {
-                // make recursive call to check if all sub-pages are valid for current language (see above)
-                $parsedMenu = $this->parseMenu($subMenuPages);
-                if (0 < count($parsedMenu)) {
-                    $pages[$index]['hasSubPages'] = true;
-                    
-                    //TODO: Remove deprecated argument in next major version
-                    $class[] = $this->arguments[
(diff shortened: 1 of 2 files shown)


Classes/ViewHelpers/Menu/AbstractMenuViewHelper.php
```diff
@@ -475,12 +475,19 @@ public function parseMenu(array $pages)
...
-                $pages[$index]['hasSubPages'] = true;
-                //TODO: Remove deprecated argument in next major version
-                $class[] = $this->arguments[
-                    $this->hasArgument('classHasSubpages') ? 'classHasSubpages' : 'classHasSubPages'
-                ];
+            // simply check if current page uid can have sub-pages
+            $subMenuPages = $this->getMenu($originalPageUid);
+            if (0 < count($subMenuPages)) {
+                // make recursive call to check if all sub-pages are valid for current language (see above)
+                $parsedMenu = $this->parseMenu($subMenuPages);
```
NamelessCoder: This needs to be handled in getMenu() somehow. If PageRepository is returning an incorrect/incomplete result it could be because our usage is incorrect or incomplete, so that is where you should look for this problem.

Even if this patch does fix the issue, I can't accept recursively calling `parseMenu` which could recursively call `getMenu` over and over - it adds potentially many more SQL queries that you can't avoid.

Which part of the `parseMenu` function is it exactly that makes the result correct? As far as I can tell there is only one thing which might change the *number* of pages and thus the count value - and that is `hidePageForLanguageUid` gets called in the loop.

If our usage of PageRepository isn't the problem then perhaps the right solution is to call this method to filter the return from `getMenu` - which would then mean that `parseMenu` instead needs to be changed to remove the call to `hidePageForLanguageUid`.

Hope you can take a look at that alternative solution! It should perform much better without adding recursion in `parseMenu` ;)

h69: Thanks for your quick response and the hint to fix this issue inside `getMenu` method.

> This needs to be handled in getMenu() somehow. If PageRepository is returning an incorrect/incomplete result it could be because our usage is incorrect or incomplete, so that is where you should look for this problem.
Sadly the core is not able to return the correct results. The own method `hidePageForLanguageUid` is necessary and handles this core issue perfectly ;)

> If our usage of PageRepository isn't the problem then perhaps the right solution is to call this method to filter the return from getMenu - which would then mean that parseMenu instead needs to be changed to remove the call to hidePageForLanguageUid
Thanks for this hint. I edited the `getMenu` method of the `PageService` class to filter the results right after `PageRepository`'s `getMenu` call and removed the `hidePageForLanguageUid` call from `parseMenu`.

NamelessCoder:
Much better - thanks for refactoring!

NamelessCoder:
@h69 Before I merge, would you mind confirming my changes (just pull from your own branch)?

h69:
@NamelessCoder looks good, thanks ;)

NamelessCoder:
First VHS contribution - celebrate with 🍰 ! ^^

PR #1450, opened 2018-02-21
Handle empty array in ExtractViewHelper

### LINKED ISSUE #1451: ExtractViewHelper: returns NULL for empty array
If the `ExtractViewHelper` is given an empty array with an arbitrary `key`, it always returns `NULL` instead of an empty array.

mbrodala:
Fixes #1451

### DIFF
--- a/Classes/ViewHelpers/Iterator/ExtractViewHelper.php
+++ b/Classes/ViewHelpers/Iterator/ExtractViewHelper.php
@@ -172,7 +172,7 @@ class ExtractViewHelper extends AbstractViewHelper implements CompilableInterfac
         if (false === is_array($iterator) && false === $iterator instanceof \Traversable) {
             throw new \Exception('Traversable object or array expected but received ' . gettype($iterator), 1515498714);
         }
-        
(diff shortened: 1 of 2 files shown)


NamelessCoder:
This looks appropriate, but changes a return type from null to array. That's okay but there appears to be at least one (more) test that needs changing.

mbrodala:
You are right, I didn't take care of the non-recursive variant.

mbrodala:
The tests are fixed now.

Classes/ViewHelpers/Iterator/ExtractViewHelper.php
```diff
@@ -172,7 +172,7 @@ protected static function recursivelyExtractKey($iterator, $key)
         if (false === is_array($iterator) && false === $iterator instanceof \Traversable) {
             throw new \Exception('Traversable object or array expected but received ' . gettype($iterator), 1515498714);
         }
-
```
NamelessCoder: Eew, how did that get in there? :)

PR #1458, opened 2018-03-15
[TASK] Allow installation of TYPO3 master

helhum:
Add replacement to typo3/cms-version in this package
as this package was removed for TYPO3 8 and merged
into typo3/cms-workspaces.

Simplify travis build by only requiring typo3/cms-core.
This is enough because all TYPO3 packages require
this package in the same version. If typo3/cms-core is required
in dev-master, all other are also installed in dev-master.

This has the additional benefit, that other direct dependencies
to TYPO3 framework packages remain unchanged, thus it is checked
by composer whether they are allowed to be installed in the version
requested for typo3/cms-core.

Add prefer-stable: true to always fetch stable versions when
possible even when composer is called without --prefer-dist.

Last but not least, replace the deprecated or operator "|"
whith the new "||" operator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,7 @@ before_script:
   - export TYPO3_PATH_WEB=$PWD/build/web
 
 install:
-  - composer require --no-update "typo3/cms-core:$TYPO3_VERSION"
-  - composer require --no-update "typo3/cms-extbase:$TYPO3_VERSION"
-  - composer require --no-update "typo3/cms-fluid:$TYPO3_VERSION"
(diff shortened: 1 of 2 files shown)


helhum:
@NamelessCoder @cedricziel Here you go and thanks for removing the dependency to `typo3/cms`. Even users for TYPO3 8.7 will benefit from that move.

Now you see the failing test with TYPO3 master.

I currently prefer not declaring compatibility with TYPO3 dev-master packages in my packages, but I see how it can simplify your testing workflow.

NamelessCoder:
Thanks a lot Helmut, both for taking the time to fix and explain this!

Issue #1462, opened 2018-03-21
random.number is not really random (no CS[P]RNG)


DanielRuf:
https://github.com/FluidTYPO3/vhs/blob/9f5166ea8df9f09f916a05f0294e3cc7781695f5/Classes/ViewHelpers/Random/NumberViewHelper.php#L88

We should use `random_int`.

cedricziel:
Nice find - care to provide a quick patch?

DanielRuf:
What is the PHP versions that should be supported at least?

I will provide a PR if I find the time.

cedricziel:
Our composer manifest says we require 7, so it's fine.

DanielRuf:
Will also change https://github.com/FluidTYPO3/vhs/blob/9f5166ea8df9f09f916a05f0294e3cc7781695f5/Classes/ViewHelpers/Random/StringViewHelper.php then as `rand` and `mt_rand` have no good seeding and `random_bytes` should be used.

cedricziel:
Please go ahead! :)

PR #1465, opened 2018-03-25
[TASK] Use random_int and random_bytes instead of rand

DanielRuf:
Unseeded `rand` and `mt_rand` do not provide real random data.

Closes https://github.com/FluidTYPO3/vhs/issues/1462

Also it is not quite clear to me why `floatval` was used. In general a number can be also decimal / numeric. That is different to a float / real / double value.


DanielRuf:
PS:

> Has built-in insurance that first character of random string is an alphabetic character (allowing safe use as DOM id for example). /

This does not seem to be the case.

DanielRuf:
The first alphabetic character was removed in https://github.com/FluidTYPO3/vhs/commit/4d167912f3a64b25a8d714d7910a8612c79c5415#diff-d477145e2d727ff5af4f5a633a394c50

PR #1469, opened 2018-03-27
[BUGFIX] Only auto-render render LanguageMenu without "as" argument

HenningCash:
Current behaviour is not like described: LanguageMenu is auto-rendered if inner content is empty and not if `as` is missing. The following example will trigger auto-rendering if there is only a single language:

```xml
<v:page.languageMenu class="list-inline" hideNotTranslated="1">
    <f:if condition="{langs -> f:count()} > 1">
        <f:then>
            <div id="nav-lang">
                <ul class="list-inline">
                    <f:for each="{langs}" as="lang">
                        <li class="{f:if(condition: lang.current, then: 'active')}">
                            <a href="{lang.url}">{lang.label}</a>
                        </li>
                    </f:for>
                </ul>
            </div>
        </f:then>
    </f:if>
</v:page.languageMenu>
```

### DIFF
--- a/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
+++ b/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
@@ -100,8 +100,7 @@ class LanguageMenuViewHelper extends AbstractTagBasedViewHelper
             'string',
             'If used, stores the menu pages as an array in a variable named according to this value and renders ' .
             'the tag content - which means automatic rendering is disabled if this attribute is used',
-            false,
-            'languageMenu'
+            false
         );
         $this->registerArgument('pageUid', 'integer', 'Optional page uid to use.', false, 0);
         $this->registerArgument('configuration', 'array', 'Additional typoLink configuration', false, []);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
This will break all implementations which expect the default `as` argument value. Those will then have to specify that explicitly to restore function.

We definitely can't do it as a `BUGFIX`, but I'm open to a change that implements this behavior - it just has to be done in every menu ViewHelper at the same time and would be tremendously helpful with a global setting which can toggle that behavior back to the original one so you can upgrade existing sites without having to refactor.

*Workaround:* by putting an HTML comment inside the `v:page.languageMenu` tag but outside the condition, you guarantee that the trimmed content is not empty, thus disabling automatic rendering. This works for all of the menu ViewHelpers!

Sketched future solution:

* AbstractMenuViewHelper has to be adjusted - it has two different variable assignment arguments which also should become optional, and there are methods like backupVariables which assumes these are not-null and must be made to ignore them if not assigned.
* All implementations subclassing AbstractMenuViewHelper need to be adjusted. Some override the render method and argument registrations (note: I'm aware this particular VH doesn't subclass that parent, but this VH should still be consistent with the other menu-type ones).
* An extension config toggle is added, default OFF, for example called "Legacy behavior for menu ViewHelper 'as' arguments", which when active changes the default value of the argument (default value can be stored in global TYPO3_CONF_VARS, different values can be assigned based on ext conf - if the assigned value is null this is the same as removing the default value). Just reference the TYPO3_CONF_VARS directly where declaring the default argument value for each VH.

So unfortunately it's not a matter of changing a default argument value. The consistency and compatibility concerns are pretty major for these ViewHelpers in particular since they are so frequently used.

HenningCash:
Okay, I haven't though about breaking BC. This PR can be closed but at least the documentation should be updated so it matches the current behavior.

Issue #1471, opened 2018-04-12
v:condition.string.isNumeric is not working as expected


jacobsenj:
Markup:

```
<v:condition.string.isNumeric value="5">
	<f:then>
		then
	</f:then>
	<f:else>
		else
	</f:else>
</v:condition.string.isNumeric>
```

Expected output:

`then`

Current output:

`else`

TYPO3 8.7.12, PHP 7.0.27

jacobsenj:
The viewhelper is using ctype_digit() to check the value. As integer values are failing the check, I would rather use is_numeric() to check the value:

```
jjacobsen@J ➜  ~  php -r "echo (ctype_digit(5) ? 'Yes' : 'NO') . PHP_EOL;"
NO
jjacobsen@J ➜  ~  php -r "echo (ctype_digit('5') ? 'Yes' : 'NO') . PHP_EOL;"
Yes
```

vs.

```
jjacobsen@J ➜  ~  php -r "echo (is_numeric(5) ? 'Yes' : 'NO') . PHP_EOL;"
Yes
jjacobsen@J ➜  ~  php -r "echo (is_numeric('5') ? 'Yes' : 'NO') . PHP_EOL;"
Yes
```

PR #1472, opened 2018-04-12
[BUGFIX] Fix condition.string.isNumeric viewhelper

jacobsenj:
The argument value type does not match the parameter type of ctype_digit() and results in a negative check for integer values passed to the viewhelper. Using is_numeric() instead of ctype_digit() fixes the problem.

Fixes: #1471


NamelessCoder:
Thanks Jens :)

Issue #1463, opened 2018-03-22
[FEATURE REQUEST] limit for ExplodeViewHelper


p2media:
It would be nice if the `ExplodeViewHelper` would have a new attribute `limit`, which corresponds to PHP's `explode()` limit parameter.

What do you think?

NamelessCoder:
Reasonable request, if anyone wants to take care of that consider it pre-approved!

PR #1473, opened 2018-04-13
[FEATURE] Added limit argument to ExplodeViewHelper

p2media:
The limit argument corresponds to the 3rd parameter "limit" of PHP's explode()-
function.

Resolves: #1463

### DIFF
--- a/Classes/ViewHelpers/Iterator/ImplodeViewHelper.php
+++ b/Classes/ViewHelpers/Iterator/ImplodeViewHelper.php
@@ -8,22 +8,33 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Iterator;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use FluidTYPO3\Vhs\Traits\ArrayConsumingViewHelperTrait;
+use FluidTYPO3\Vhs\Traits\TemplateVariableViewHelperTrait;
 use TYPO3\CMS\Fluid\Core\ViewHelper\Facets\CompilableInterface;
-use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithContentArgumentAndRenderStatic;
+use TYPO3\CMS\Fluid\Core\Rendering\RenderingContextInterface;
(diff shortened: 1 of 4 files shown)


p2media:
I need to update the Unit Tests later to check if the ViewHelpers are working properly 😉

NamelessCoder:
Instead of this, it's better to split these two ViewHelpers up now and simply call either `implode()` or `explode()` directly and not share their arguments. It is in particular very bad practice to register and then remove a ViewHelper argument so you should definitely avoid doing that in your projects.

p2media:
I split those ViewHelpers. But now, the `resolveGlue()` function is the same in both ViewHelpers. Should we leave this? 🤔

p2media:
Hi,
I just wanted to ask, If you had time to check my changes

Classes/ViewHelpers/Iterator/ImplodeViewHelper.php
```diff
@@ -33,6 +44,61 @@ class ImplodeViewHelper extends ExplodeViewHelper implements CompilableInterface
...
+     *
+     * @param array $arguments
+     * @return string
+     */
+    protected static function resolveGlue(array $arguments)
+    {
+        $glue = $arguments['glue'];
+        if (false !== mb_strpos($glue, ':') && 1 < mb_strlen($glue)) {
+            // glue contains a special type identifier, resolve the actual glue
+            list ($type, $value) = explode(':', $glue);
```
NamelessCoder: Note to self: at some point this should be removed (causing a breaking change). Intention in future is to use `v:const` (which may indeed become `f:const` because constants are universally useful; or perhaps even integrate constants output natively with something like `{FOOBAR}` if a variable named `FOOBAR` isn't set.

NamelessCoder:
Merge-privileged people: feel free to merge if you can confirm with a quick local test that `limit` works!

NamelessCoder:
Merged manually - thanks for the patch @p2media!

Issue #1486, opened 2018-06-18
v:format.json.encode converts "false" to "{}" on top level


maechler:
The current implementation of `v:format.json.encode` converts the value `false` to `{}`, when used on top level. Although the JSON spec allows any value (object, array, true, false, null, numbers) to be on top level. (https://stackoverflow.com/questions/17220924/does-the-top-level-value-in-json-have-to-be-an-array-or-object)

`{v:format.json.encode(value: falseVariable)} --> returns {}`
`{v:format.json.encode(value: trueVariable)} --> returns true`

The problem is the following code because `empty(false)` evaluates to `true`: 
https://github.com/FluidTYPO3/vhs/blob/ed6b9219f2f8fd0439fbffd0f72b26ee6159ab70/Classes/ViewHelpers/Format/Json/EncodeViewHelper.php#L100-L102

If you think this could safely be removed, I could prepare a pull request.

NamelessCoder:
This sounds like it is safe to remove. PR would be appreciated!

PR #1487, opened 2018-06-19
[BUGFIX] Remove return {} on empty value in order to allow false and null values (#1486)

maechler:
See #1486.


NamelessCoder:
The errors are not related. Once fixed I will update your PR branch and it should turn green. You don't need to do anything. In fact, the PR looks perfect and the unit test changes are more than we usually ask for ;)

maechler:
@NamelessCoder Great! Thanks =)

NamelessCoder:
Thanks for the patch @maechler - very nicely done for a first contribution!

Issue #1491, opened 2018-06-25
Migrating from TYPO3_DB


rafu1987:
There are still some occurrences of old TYPO3_DB, but the extension is marked to work with TYPO3 9.3.x. Either remove the TYPO3_DB calls or add the `typo3db_legacy` extension as a dependency.

https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Database/Migration/Index.html
https://extensions.typo3.org/extension/typo3db_legacy/

cedricziel:
Aye, would you mind creating a pull request for that?

NamelessCoder:
> Aye, would you mind creating a pull request for that?

And by "that" we'd absolutely prefer a refactoring of the limited number of places where we currently don't use Doctrine ;)

rafu1987:
I will look into it. I think, first it should be made compatible with TYPO3 9.3.x, as currently it isn't and then think about refactoring (however this would be implemented).

NamelessCoder:
The way you're solving this is actually precisely what I meant with "refactoring" so: excellent work, it is greatly appreciated if you continue in this direction!

rafu1987:
@NamelessCoder thanks you very much :-)
I created a pull request and removed the class check. Also, I removed the class check here in this file, too: `Classes/ViewHelpers/Content/AbstractContentViewHelper.php`

I added the doctrine stuff here and fixed #1433 along with it: `Classes/ViewHelpers/Resource/AbstractResourceViewHelper.php`

And added the doctrine stuff in this file as well: `Classes/ViewHelpers/Resource/Record/FalViewHelper.php`

Now, all the TYPO3_DB calls should be replaced and I hope everything still works as expected :-)

NamelessCoder:
@glucka For https://github.com/FluidTYPO3/vhs/pull/1492 the main missing thing is to use named parameters in the doctrine queries as opposed to the raw value comparisons. Aside from that there could be other occurrences that also need to be taken care of.

You can definitely help to complete it. But in order to do that there's a little something that has to be officially taken care of ;)

*I've added you to the team of FluidTYPO3 maintainers!*

Welcome on board and thank you so much for all the help. If/when you accept the invitation you will gain access to pushing updates to the PR branch on the users' forks and that's a good way to help push stalled PRs along.

I have only a few "rules" you should keep in mind:

* Please don't push to development, staging or master branches directly for the time being.
[...]

PR #1492, opened 2018-06-25
[BUGFIX] Replace TYPO3_DB with doctrine

rafu1987:
Some replacements for #1491

### DIFF
--- a/Classes/ViewHelpers/Condition/Page/IsLanguageViewHelper.php
+++ b/Classes/ViewHelpers/Condition/Page/IsLanguageViewHelper.php
@@ -8,6 +8,9 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Condition\Page;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Database\Query\QueryBuilder;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractConditionViewHelper;
 
 /**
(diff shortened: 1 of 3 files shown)


Classes/ViewHelpers/Condition/Page/IsLanguageViewHelper.php
```diff
@@ -44,7 +47,22 @@ protected static function evaluateCondition($arguments = null)
         if (true === is_numeric($language)) {
             $languageUid = intval($language);
         } else {
-            $row = $GLOBALS['TYPO3_DB']->exec_SELECTgetSingleRow('uid', 'sys_language', "title='" . $language . "'");
+            if (class_exists(ConnectionPool::class)) {
```
NamelessCoder: No need to check for the class existing - we have 8.7 as minimum so it is guaranteed to exist. Please remove the check and the code in the "else" block which calls TYPO3_DB as well. We only need the doctrine version now :)

NamelessCoder:
Thanks for the patch @rafu1987 - this is the right direction. Whitespace formatting and a few calls to doctrine need to be changed but I will happily handle those if you remove the conditions as mentioned.

Classes/ViewHelpers/Condition/Page/IsLanguageViewHelper.php
```diff
@@ -44,7 +47,18 @@ protected static function evaluateCondition($arguments = null)
...
         } else {
-            $row = $GLOBALS['TYPO3_DB']->exec_SELECTgetSingleRow('uid', 'sys_language', "title='" . $language . "'");
+            /** @var QueryBuilder $queryBuilder */
+            $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable('sys_language');
+
+            $row = $queryBuilder
+                ->select('uid')
+                ->from('sys_language')
+                ->where(
+                    $queryBuilder->expr()->eq('title', '"' . $language . '"')
```
cedricziel: General nitpick: We should use named parameters for the values, shouldnt we?

glucka: in this case:
```$queryBuilder->expr()->eq('title', $queryBuilder->createNamedParameter($language, \PDO::PARAM_STR))```
right?

Comment (NamelessCoder) [reply Classes/ViewHelpers/Condition/Page/IsLanguageViewHelper.php]: Indeed - createdNamedParameter with *both* arguments even when second argument matches default argument value, is our preferred way :)

Comment (glucka) [reply Classes/ViewHelpers/Condition/Page/IsLanguageViewHelper.php]: Ok, then i make so every query, ok?
```
            $queryBuilder->createNamedParameter($language, \PDO::PARAM_STR, ':title');

            $row = $queryBuilder
                ->select('uid')
                ->from('sys_language')
                ->where(
                    $queryBuilder->expr()->eq('title', ':title')
                )
                ->execute()
                ->fetch();
```

NamelessCoder: Also fine for me. Personally I prefer the way of just passing `$queryBuilder->createNamedParameter($language, \PDO::PARAM_STR)` into the `->eq()` or whatever, rather than actually naming the parameters - but that's just taste ;)

dmitryd:
It would be nice to have this merged because EXT:vhs cannot be used with v9.5 natively now. Making the code perfect is good but can be done after the fix allows people to run the ext and test it... Just an opinion :)

dmitryd:
@NamelessCoder Thank you a lot, this moves our projects forward :)

PR #1495, opened 2018-06-27
[FEATURE] Add preloadCss option to use with loadCSS

rafu1987:
**loadCSS**
https://github.com/filamentgroup/loadCSS


NamelessCoder:
Asset feature removed in upcoming version.

PR #1496, opened 2018-06-27
[FEATURE] Added options compressCss and compressJs

rafu1987:
Added options compressCss and compressJs to compress CSS and JS output with **matthiasmullie/minify** in non-composer mode

https://github.com/matthiasmullie/minify


cedricziel:
I'm sorry, but we cannot accept your pull-request.

* we don't embed other libraries
* we cannot upload an extension to TER that embeds non-GPLv2 code

A PR to add a Signal to process a chunk in a 3rd party extension would be greatly appreciated.

Thank you for your time and effort!

cedricziel:
By the way: Maybe we can use the hooks provided by the core, by calling `minifyJavascript` on `GeneralUtility`.

https://github.com/TYPO3/TYPO3.CMS/blob/b5cd555659d44a2fee11af77f9634ab1ac0599dd/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L1756-L1774

rafu1987:
Ah that's too bad but I kind of expected that.
So if I was to build an extension that embeds the minify class, you could define that as a dependency in vhs and I would add a signal to that extension?

rafu1987:
I thought about implementing the core functionality of `compressJs` and `compressCss` but there are some drawbacks there, as new files are generated in `typo3temp/assets/compressed` and we would have to rewrite a lot of stuff in `AssetService.php` which was not really what I was hoping for.

https://github.com/TYPO3/TYPO3.CMS/blob/62de88be908950845962431843053060ecc5d3d7/typo3/sysext/core/Classes/Resource/ResourceCompressor.php

cedricziel:
Yes, I've been there before :)

My thoughts:
* add a signal in VHS that returns the processed chunk
* write that to disk
* you can then create a new extension, depending on VHS, slotting the signal
* process whatever needs to be processed on your end

rafu1987:
See #1497

PR #1497, opened 2018-06-27
[FEATURE] Add signal slot when writing asset files

rafu1987:
vhsminify extension will shortly be up in TER

**This is the processing in the extension:**
https://bitbucket.org/rafu1987/vhsminify/src/master/Classes/Hooks/Minify.php


Classes/Service/AssetService.php
```diff
@@ -763,6 +765,13 @@ public function clearCacheCommand($parameters)
      */
     protected function writeFile($file, $contents)
     {
+        // Minify assets with vhsminify
+        if (ExtensionManagementUtility::isLoaded('vhsminify')) {
```
cedricziel: Please do not check for loaded 3rd party extensions, the signal, if any, should be available to any 3rd party.

Please use a constant for the signal name.

Classes/Service/AssetService.php
```diff
@@ -61,6 +61,11 @@ class AssetService implements SingletonInterface
      */
     protected static $cacheCleared = false;
 
+    /**
+     * @var string
+     */
+    protected $minifySignal = 'minifyAssets';
```
cedricziel: the field should be a constant so it's actually accessible in other peoples' localconf

rafu1987:
Extension is now online in TER
https://extensions.typo3.org/extension/vhsminify/

I will test it in TYPO3 9.3.x later and add support.

Classes/Service/AssetService.php
rafu1987: I'm sorry, I don't know what you mean by that.

cedricziel: Currently the field is protected and a member of the class. - Which means a user that wants to slot the signal cant use it.

Please use

```

Comment (rafu1987) [reply Classes/Service/AssetService.php]: That's weird, because I was able to slot it just fine in my extension vhsminify. I changed it as you suggested though :-)

Comment (cedricziel) [reply Classes/Service/AssetService.php]: Thx. Yeah, you're slotting it with the constants string value, not with the constant itself :)

https://bitbucket.org/rafu1987/vhsminify/src/16a42770f8f8a430cb26d851c114a3efa6cb8554/ext_localconf.php?at=master&fileviewer=file-view-default#ext_localconf.php-25

I suggest you use the class constant and the signal name constant to register now:

```php
        $signalSlotDispatcher->connect(
            \FluidTYPO3\Vhs\Service\AssetService::class,
            \FluidTYPO3\Vhs\Service\AssetService::MINIFY_SIGNAL,
            'RZ\Vhsminify\Hooks\Minify',
            'minifyAssets',
            false
        );
```

That way you'll notice immediately if bad things happen.

rafu1987: Okay, thanks for the explanation :) First time for me, adding a signal slot. Great, that it works now :-)

Classes/Service/AssetService.php
```diff
@@ -29,6 +31,11 @@
 class AssetService implements SingletonInterface
 {
 
+    /**
+     * @var string
+     */
+    const MINIFY_SIGNAL = 'minifyAssets';
```
NamelessCoder: This should be a completely generic name like "writeAssetFile" - a subscriber can do literally anything to the source and we don't want to make it seem like it's only for minifying sources.

You may also want to pass `&$file` by reference when calling the signal - and make the method return $file. Then where `writeFile()` is called, use the return value as filename. This would then allow the signal subscriber to change the target file location as well. A companion change which passes the asset chunk settings to `writeFile` as a third parameter would also be acceptable (but make it default to `null` so the signature remains compatible and we can sneak it in as a non-breaking patch). I guess those would be quite handy to receive in the signal as well.

rafu1987: I agree with you on the name change :-)

I changed the function to the following, is that correct?

```
    /**
     * @param string $file
     * @param string $contents
     */
    protected function writeFile($file, $contents)
    {
        /** @var Dispatcher $signalSlotDispatcher */
        $signalSlotDispatcher = GeneralUtility::makeInstance(Dispatcher::class);
        $signalSlotDispatcher->dispatch(__CLASS__, static::ASSET_SIGNAL, [&$file, &$contents]);

        GeneralUtility::writeFile($file, $contents, true);
    }
```

I had to return `$file` and `$contents` as an array in my method, otherwise I would get the following error: `The slot method RZ\Vhsminify\Hooks\Minify->minifyAssets()'s return value is of an not allowed type (string)` or if I pass an array only with `$file` `The slot method RZ\Vhsminify\Hooks\Minify->minifyAssets() returned a different number (1) of arguments, than it received (2).`

I didn't need to change the `$file` variable in the `writeFile` function though.

And finally, I don't know how to pass the asset chunk to the `writeFile` function (I think it's called twice in `AssetService.php`). Maybe you could point that out to me. :-)

NamelessCoder:
It will take a while before I can come back to this one - sorry. Other pressing priorities at the moment. I think the idea with the signal slots is that we don't pass values by reference but instead overwrite them with what is returned from the signal slot dispatcher (hence the requirement for number of returned array values). And what I meant with refactoring was that inside AssetService, if we make writeFile return the filename and contents and use those values, the code that executes after files are written will also see that the filename/content was changed.

Hope that makes sense - if not, just leave it here and I'll continue this patch when I get the time.

PR #1499, opened 2018-07-11
[TASK] Add option to disable translation fallback in menus

### LINKED ISSUE #1498: MenuViewHelper does not hide untranslated pages
We have set up our site with the following to ensure no default content is displayed on translated pages:

```
config {
  sys_language_overlay = 0
  sys_language_mode = strict
}
```

However, the `MenuViewHelper` does not respect this and in fact always falls back to including pages from the default language in the menu.

This is caused by `PageService::getMenu()` which filters the menu pages by `PageService::hidePageForLanguageUid()` but does not offer an option to disable the `$normalWhenNoLanguage` flag.

mbrodala:
Fixes #1498

### DIFF
--- a/Classes/Service/PageService.php
+++ b/Classes/Service/PageService.php
@@ -52,6 +52,7 @@ class PageService implements SingletonInterface
      * @param boolean $includeNotInMenu
      * @param boolean $includeMenuSeparator
      * @param boolean $disableGroupAccessCheck
+     * @param boolean $normalWhenNoLanguage
      *
      * @return array
      */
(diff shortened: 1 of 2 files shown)


mbrodala:
(Build failure is unrelated.)

neufeind:
@mbrodala Can you do a rebase of this?

mbrodala:
Done.

Classes/Service/PageService.php
```diff
@@ -60,7 +61,8 @@ public function getMenu(
         array $excludePages = [],
         $includeNotInMenu = false,
         $includeMenuSeparator = false,
-        $disableGroupAccessCheck = false
+        $disableGroupAccessCheck = false,
+        $normalWhenNoLanguage = false
```
NamelessCoder: We have a couple of other patches which all try to change the method signature for getMenu which in turn affects any ViewHelper using menu functions. I'm extremely hesitant to change this bit unless it gets changed to a way that will prevent future signature changes, which means either:

* Implementing a DTO that contains menu configuration instructions.
* Or replacing this with whichever core method allows us to read a menu without rendering it as HTML.

But I don't think we should open the flood gates for this signature change because it's going to happen again and again, and the signature will become bloated. It's already pretty bad...

NamelessCoder (changes requested):
We need a strategy that moves us around the issue of signature changes for menu-related methods...

PR #1514, opened 2018-09-12
[FEATURE] Support nesting in v:variable.set

pkerling:
SetViewHelper could only set direct child properties of arrays or
objects before, which is inconvenient when working with nested
structures. This change adds code to access properties nested at
arbitrary levels inside mixed array/object hierarchies.

This is actually much more complicated then I'd like, but so far I haven't found an easy way around it. Suggestions to make it more readable very welcome. Maybe there's already a helper function for this somewhere that I missed?

### DIFF
--- a/Classes/ViewHelpers/Variable/SetViewHelper.php
+++ b/Classes/ViewHelpers/Variable/SetViewHelper.php
@@ -91,21 +91,91 @@ class SetViewHelper extends AbstractViewHelper implements CompilableInterface
                 $variableProvider->remove($name);
             }
             $variableProvider->add($name, $value);
-        } elseif (1 === mb_substr_count($name, '.')) {
+        } else {
             $parts = explode('.', $name);
             $objectName = array_shift($parts);
-            $path = implode('.', $parts);
             if (false === $variableProvider->exists($objectName)) {
                 return null;
             }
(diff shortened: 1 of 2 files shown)


NamelessCoder:
I think we should solve this in Fluid instead, making it possible to `assign('variable.with.path', 'value')`. This could be done in VariableProvider´s `add()` method by checking if `$identifier` contains a dot and if it does, https://github.com/NamelessCoder/typo3-cms-data-processors/blob/master/Classes/StructuredVariablesProcessor.php#L82 should give some inspiration how we could handle that one (and by the way... implementing this in Fluid makes that package obsolete).

The code on the link uses ObjectAccess and obviously Fluid can't do that - instead it can use the internal methods on VariableProvider that resolves a value. We can change those methods so they return references to avoid the deference-if-parent-is-array problem described in the comment.

Then after all that we should simply deprecate the VHS ViewHelper.

NamelessCoder:
Like this: https://github.com/TYPO3/Fluid/pull/425 ;)

pkerling:
As long as it works and the feature set is roughly the same as with `vhs` - sure :-) Thanks for the PR.

NamelessCoder:
Closing this one, will pursue getting the Fluid-native feature merged.

PR #1516, opened 2018-09-18
[BUGFIX] Do not render asset ViewHelpers twice

pkerling:
The asset content currently gets rendered
1. when the fluid template containing the asset is rendered and
2. when assets are built

This is not necessary and might produce wrong results because the asset
should really only be rendered in the context of the template it is used
in and not at some other time when the context might be different,
required variables not available etc.

Fix by using the content saved during the initial render instead of
re-rendering.


pkerling:
Thanks!

PR #1517, opened 2018-09-30
[TASK] Add fail-allowed builds for notice and deprecations

NamelessCoder:
Adds two new error-silenced builds that allow seeing E_NOTICE and E_USER_DEPRECATED warnings separately, based on TYPO3 dev-master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,29 @@ php:
 
 matrix:
   include:
-    - php: "7.2"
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Don't merge - will self-merge when resolved.

cedricziel:
Unfortunately, the checkbox is only good If you keep the PR open in a foreground tab.

cedricziel:
Blinded by the green. The famous earth-band song :)

PR #1519, opened 2018-10-09
Add named parameters to db querys

glucka:


Classes/ViewHelpers/Resource/AbstractResourceViewHelper.php
```diff
@@ -92,18 +95,30 @@ public function getFiles($onlyProperties = false, $identifier = null, $categorie
...
-            );
+                ->andWhere(
+                    $queryBuilder->expr()->in('uid_local', ':categories')
+                )
+                ->execute()
+                ->fetchAll();
 
             $fileUids = array_unique(array_column($rows, 'uid_foreign'));
 
+            // Now FIX UIDs if they are sys_file_metadata UIDs
```
NamelessCoder: This looks like maybe it fixes a bug or adds new logic - can you write a few words about why the references are now resolved to original UIDs? This would, if I still remember things right, change the type of value you get returned and might cause errors for example if you call `f:image` with `treatIdAsReference` and expect the UID values to be reference UIDs so I'm not sure it's safe / related to doctrine switch?

glucka:
Thanks for the url @NamelessCoder! Installed just now ;)

Classes/ViewHelpers/Resource/AbstractResourceViewHelper.php
NamelessCoder: Thanks @rafu1987 that explains it :) @glucka let's separate this fix into a PR by itself so it gets included in the changelog!

NamelessCoder:
Feels good to get rid of these :)

@glucka would you mind following up with a PR for https://github.com/FluidTYPO3/vhs/pull/1519/files#diff-483bb1e8b46b9340f483e0bc253bab7aR143 which replaces the SQL calls with usage of I think it's ResourceFactory::getFileReferences, which should return the same kind of data? I suspect that entire method can be replaced with a call to fetch references and maybe a loop to process them.

glucka:
OK I will try ;)

glucka:
https://github.com/FluidTYPO3/vhs/pull/1519/files#diff-483bb1e8b46b9340f483e0bc253bab7aL131
In which case can $GLOBALS['TSFE']->sys_page be empty? I don't really understand this part...

NamelessCoder:
That looks like a VERY old way of testing if we're in FE and can use the API via TSFE->sys_page to read references. But if I'm right in what I suspect, I think we can completely drop this as well and replace it with the method that uses ResourceFactory to read references - which then may need special arguments to respect or ignore the hidden/delete etc. status based on preview mode or not (preview mode can be checked on TSFE and there are some cases in I think the content:get VH that does this).

If we can use one method that works in FE and BE alike and doesn't require SQL statements then that's our hero method ;)

glucka:
I think we cannot use ResourceFactory::getFileReferenceObject because this method for BE use the TYPO3\CMS\Core\Database\RelationHandler and this require valid DB fields (TCA Configs)
I don't see a better solution at this time. :(

glucka:
The main problem is to find out the $referenceUid in BE Mode. 
In FE works fine over DB querys in $this->fileRepository->findByRelation($table, $field, $sqlRecordUid);
But in BE Mode Typo3 use the RelationHandler and this need valid DB fields.
I think it will never works with flexform fields.

glucka:
In this case is possibly cleaner if we remove the FE method (findByRelation) and use for both native DB queries?

tobiasgraeber:
I think commit  [BUGFIX] Replace TYPO3_DB with Doctrine usage (#1519) caused side effects. For example, please see: https://github.com/FluidTYPO3/vhs/issues/1524

In short:
Vviewing "disabled" pages in typo3 as an page editor in "preview" mode causes now a fatal error ...

PR #1523, opened 2018-10-17
[BUGFIX] Fix #1504: use "pages" instead of "pages_language_overlay" since TYPO3 9.5

### LINKED ISSUE #1504: LanguageMenuViewHelper still selects records from pages_language_overlay table
Hello, 
I am trying to use the `<v:page.languageMenu>` in TYPO3 v9 and receive this error: 

```
There is no entry in the $TCA array for the table "pages_language_overlay". This means that the function enableFields() is called with an invalid table name as argument
```

I've noticed, that the table `pages_language_overlay` is still in use. 
https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php#L317

I guess that must be changed for compatibility to v9.

dmitryd:
This fixes an exception about missing TCA definition for `pages_language_overlay` in TYPO3 9.5.0+ while maintaining compatibility with lower TYPO3 versions.

Close: #1504

### DIFF
--- a/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
+++ b/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
@@ -313,8 +313,8 @@ class LanguageMenuViewHelper extends AbstractTagBasedViewHelper
             }
         }
 
-        // Select all pages_language_overlay records on the current page. Each represents a possibility for a language.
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks Dmitry!

Issue #1448, opened 2018-02-19
language/country flags changed from SVG to PNG in typo3 8.x


SvenMb:
In file  vhs/Classes/Utility/CoreUtility.php the path for getLanguageFlagIconPath() is still set to 'Resources/Public/Icons/Flags/SVG/', but should be 'Resources/Public/Icons/Flags/PNG/' since with the typo3 8.7 they choosed to remove the SVG flags.

Also the default for 'flagImageType' is svg instead of png.

regards
Sven

AMartinNo1:
I can confirm this and in the current master (26/04/18) the last directory "PNG" has been dropped.

https://github.com/TYPO3/TYPO3.CMS/tree/master/typo3/sysext/core/Resources/Public/Icons/Flags

PR #1526, opened 2018-10-18
[TASK] Country flags changed from SVG to PNG

ayacoo:
Related #1448

### DIFF
--- a/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
+++ b/Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
@@ -86,7 +86,7 @@ class LanguageMenuViewHelper extends AbstractTagBasedViewHelper
         );
         $this->registerArgument('useCHash', 'boolean', 'Use cHash for typolink', false, true);
         $this->registerArgument('flagPath', 'string', 'Overwrites the path to the flag folder', false, '');
-        $this->registerArgument('flagImageType', 'string', 'Sets type of flag image: png, gif, jpeg', false, 'png');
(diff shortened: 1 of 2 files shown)


Classes/ViewHelpers/Page/LanguageMenuViewHelper.php
```diff
@@ -86,7 +86,7 @@ public function initializeArguments()
         );
         $this->registerArgument('useCHash', 'boolean', 'Use cHash for typolink', false, true);
         $this->registerArgument('flagPath', 'string', 'Overwrites the path to the flag folder', false, '');
-        $this->registerArgument('flagImageType', 'string', 'Sets type of flag image: png, gif, jpeg', false, 'svg');
+        $this->registerArgument('flagImageType', 'string', 'Sets type of flag image: png, gif, jpeg', false, 'png');
```
NamelessCoder: We must not change the default value of parameters since that constitutes a breaking change. The value further isn't the same on different platforms which means we have to leave this up to the developer creating the template. To preserve compatibility we leave this as `svg` and on 8.7 you then must manually opt-in to using the PNG format.

The other part of this patch is of course still required!

NamelessCoder:
Thanks for the patch @ayacoo - a tip for future patches: if you create a new branch starting from `development` and then publish that and make the PR from there, you preserve your ability to pull to your local development branch without causing problems after the PR branch is merged. We usually squash and merge which means your commit on `development` is different from what comes from our `development` branch and you get a conflict.

ayacoo:
Thanks for your hints :-)

PR #1533, opened 2018-10-24
[TASK] Allow to use f:resource path and improve error message on invalid path

BenjaminBeck:


NamelessCoder (changes requested):
We have to consider this very, very carefully as it would allow passthrough to public of any file that can be read. The following also has to be taken into consideration:

* The end goal is to replace the VHS assets
* They are intended to work in TS as well as in Fluid so changes probably should not be done in VH classes
* The reason you get this path is likely because of `absRefPrefix` or other
* There are ways to refer to a file in a specific extension's resources folder by `FILE:EXT:` reference
* It is equally possible to mark the asset as "external" which prevents the absolute path resolving, and as "standalone" which prevents it from being read by PHP

Also a small tip: checking the first character of a string is much easier if you simply compare the strings. Any number of methods will do: `if ($string{0} === $char)` - `if (substr($string, 0, 1) === $char)` or `strncmp($string, $char, 1)`. The methods you should not use would include preg matching and strpos/stripos, both of which are far more expensive as they will analyse the entire string instead of an explicit, single character. See for example https://review.typo3.org/#/c/58533/ which fixes this problem in the TYPO3 core.

BenjaminBeck:
> The bad part is that introducing new exceptions means the next release is a breaking release. 

I didn´t consider this. But the error I got was also a exception (Screenshot).
So we would basically check for the same condition but ealier and give a more usefull exception.
The checks for ``$errorLevel`` and ``$this->exceptionalErrors`` have to be added.
But i am not 100% sure that this might nevertheless be a breaking release.

Is there a play when such a release will happen? Maybe we put this asisde until then?

> Regarding RFC

My personal plan/wishlist for handling assets does contain things like es6/typescript transpilation, node modules and lazy module loading. So I will at some time switch to a docker based build system in my development environment. AFAIKT this would use babel or webpack. 
The resulting file would still need to somehown go into TYPO3. I saw that the core now also uses TypeScript for its JavaScript, the might already have something similar..

2018-10-25 14_41_57-typo3 exception

NamelessCoder:
> Is there a plan when such a release will happen?

No solid plan but I would at the very least expect it to not happen until it absolutely has to. Since we only release in the most recent major version it means forcing every user to adopt the breaking change or fall behind on bug fixes etc.

> Maybe we put this asisde until then?

Definitely, we'll just leave it open - it also helps when users can find these by searching, especially when the thread contains plenty workarounds like this one does ;)

> My personal plan/wishlist for handling assets does contain things like es6/typescript transpilation, node modules and lazy module loading.

My hope is that there is a package that provides the API to register and dependency-order such assets - and has an option for plugins to be added that will do the things you mention, or does them natively.

I believe the TYPO3 core's current TypeScript implementation is a build-time feature, not a feature of TYPO3 itself. Even so, it appears to me to be a slightly too opinionated thing to just say "this is your only option in FE". For our use case here I think we can safely ignore TypeScript and aim for a much more generic solution that handles arbitrary assets and supports plugins.

Issue #1443, opened 2018-02-01
v:media.video preload doesn't work


Williwurstig:
If you set the argument preload, the output is preload="preload".
`<v:media.video src="{image.originalFile.publicUrl}" preload="none"
                                                   width="1150" height="476"`
output: 
`<video width="1150" height="476" preload="preload"><source src="Headervideo_Farbauflo__sung_kl.mp4" type="video/mp4" /></video>`

VideoViewHelper.php Line 175:
`$tagAttributes['preload'] = 'preload'`
must be changed to:
`$tagAttributes['preload'] = $this->arguments['preload'];`

neufeind:
For video that is already fixed. Duplicate of #1308.
Unresolved for audio.

PR #1536, opened 2018-10-30
[BUGFIX] Use preload argument value for f:media.audio

neufeind:
Fixes: #1443


NamelessCoder:
@neufeind with the patch to save the day :) thanks for the time you've spent on FluidTYPO3 projects today Stefan!

Issue #1440, opened 2018-01-26
TagViewHelper doesn't trim its content


jdoubleu:
I was wondering whether it was intentional, that the actual content inside a rendered tag (see https://github.com/FluidTYPO3/vhs/blob/development/Classes/Traits/TagViewHelperTrait.php#L130) should not be trimmed.

In the TagViewHelper the trimmed content will be returned if no tag name is provided (compare https://github.com/FluidTYPO3/vhs/blob/development/Classes/Traits/TagViewHelperTrait.php#L124).
However this trimmed content isn't used for the actual tag content.

I'm going to open a PR but wasn't sure if there is any reason why this viewhelper behaves like it does now.

PR #1537, opened 2018-10-30
[TASK] For a TagViewHelper use the trimmed content

neufeind:
The trimmed content was used when the tagname was empty,
but if there was a tag the content was not trimmed.
That should be unified.

Fixes: #1440


NamelessCoder:
Re-labeled to bugfix since it kind of is, and will appear in change log this way. Thanks @neufeind!

PR #1550, opened 2018-11-27
[BUGFIX] Fix for inactive languages

advOpk:
While being sure that using getPagesOverlay() is not the optimal way of verifying if the page is translated into a particular language it is still a native way.


dmitryd:
Probably obsolete by #1556

NamelessCoder:
Obsolete after #1556 - sorry about the delay @advOpk, the other patch also solves an additional issue we had with the language menu so I merged that instead.

PR #1569, opened 2019-02-08
[BUGFIX] Fix invalid link for shortcuts when useShortcutTarget is enabled

hahrens:
The parseMenu-Method uses the old original $page array which is modified when useShortcutTarget is enabled. So the modified data in $pages[$index] should be used to get the correct link for the shortcut target.


NamelessCoder:
download (6)

YOUR NICE, PROPERLY FORMATTED DATA FILE IS ACCEPTABLE!

Issue #1579, opened 2019-04-02
UnlessViewHelper is not compatible with typo3fluid/fluid v2.6.0


orion828:
In Typo3/Fluid v2.6.0 was removed automatic registration of `condition` argument from AbstractConditionViewHelper (link: https://github.com/TYPO3/Fluid/commit/a67b31f9e6ecb015d0f47892fce46cf64110fd15). Now there is error when you trying to use `Vhs\UnlessViewHelper` with new version of Fluid.

PR #1580, opened 2019-04-02
[BUGFIX] Fix for #1579. Add argument  to UnlessViewHelper

orion828:
Automatic registration of condition argument was removed from AbstractConditionViewHelper since Typo3/Fluid v2.6.0


orion828:
Automatic registration of `condition` argument was removed from `AbstractConditionViewHelper` since `Typo3/Fluid v2.6.0`

Issue #1574, opened 2019-03-14
v:resource.record.fal throws Exception if page in rootline is disabled


biedert:
When using the `v:page.resources.fal` Viewhelper with `slide="-1"`, an Error occurs, if one of the pages in the rootline is disabled:

`Argument 3 passed to TYPO3\CMS\Frontend\Page\PageRepository::getFileReferences() must be of the type array, boolean given, called in     #####typo3conf/ext/vhs/Classes/ViewHelpers/Resource/Record/FalViewHelper.php`

This ist because the same method (`FluidTYPO3\Vhs\ViewHelpers\Resource\Record\AbstractRecordResourceViewHelper->getRecord()`) is used to fetch the current page record AND each page record in it's rootline, which results in an empty record for the disabled page which leads to the exception. A possible solution would be to also remove the Hidden-Restriction in the method if the current page uid is not equal to the page uid it is trying to fetch.

biedert:
Possible bugfix: #1573

hansup:
I encountered the same issue, if I use the viewhelper `v:page.resources.fal` and a page in the rootline is hidden. If I apply he bugfix #1573 it is solved, but ... in my case the image of the hidden page is rendered, which is not desired behaviour.

What solves the problem in a desired way is casting the empty $record variable to an array:
File: 
`line:111 /vhs/Classes/ViewHelpers/Resource/Record/FalViewHelper.php`

Change:
`$fileObjects = $this->pageRepository->getFileReferences($table, $field, $record);`
to
`$fileObjects = $this->pageRepository->getFileReferences($table, $field, (array)$record);`

fnagel:
Is it possible this issue is back again with TYPO3 9.5 and VHS 5.2? My exception after updating from TYPo3 8.7 and VHS 4.4.0:

```
Argument 3 passed to TYPO3\CMS\Frontend\Page\PageRepository::getFileReferences() must be of the type array, boolean given, called in /var/www/htdocs/typo3conf/ext/vhs/Classes/ViewHelpers/Resource/Record/FalViewHelper.php on line 111
```

See here for more complete stack trace: https://pastebin.com/wtSerb6Y

fnagel:
Update: works with 1be323f5530919f17dc51af564dc8d16becdf4d2 but is broken with TER 5.2.0

PR #1582, opened 2019-04-18
[BUGFIX] Fixes fatal error when sliding up the rootline over hidden page

hansup:
Bugfix for #1574


NamelessCoder:
Thanks for the PR @hansup - much appreciated! I ended up solving it with a guard clause instead to prevent attempts to fetch records when the origin record is invalid/unavailable.

PR #1592, opened 2019-09-11
wrong database request

123AltmarkAgentur:
if variable "language" is set in the viewhelper for example "0,1", nothing return.


NamelessCoder:
Solved by https://github.com/FluidTYPO3/vhs/pull/1594

PR #1594, opened 2019-09-12
[BUGFIX] Use keyword "root" for default pidInList in language mode

123AltmarkAgentur:
if variable "language" is set in the viewhelper for example "0,1", nothing return.


NamelessCoder:
Hi @123AltmarkAgentur,

Thanks for the patch - which version of TYPO3 did you use when creating the patch?

123AltmarkAgentur:
Hi @NamelessCoder,

I'm using Typo3 8.7.27.. Thank You!

PR #1597, opened 2019-09-19
[BUGFIX] Fix OnceViewHelper

### LINKED ISSUE #1544: Once.Cookie ViewHelper doesn't work
The `Once.Cookie` ViewHelper doesn't work at all. The first the the page is rendered, the content is missing, and then it is shown all the time. Almost backwards, but that the content is missing initially is not user-dependent.

Looking at the code, I see no way that `storeIdentifier()` would be called, which fits the fact that I never get any cookie. But there are other quirks as well, for example `removeIfExpired` removes the cookie unconditionally, and `removeCookie` doesn't just unset the cookie, but also tries to unset a field in the `$_SESSION` array using the same identifier, which seems bad.

Jan1989:
Added storeIdentifier to evaluateCondition function in AbstractOnceViewHelper.php
Removed removeCookie and removeIfExpired function from CookieViewHelper, cookies should be deleted by the browser automatically after expiring.

Should fix #1544


NamelessCoder:
Processed locally and merged.

PR #1611, opened 2019-11-26
[BUGFIX] Allow sorting by DateTimeInterface

### LINKED ISSUE #1610: v:iterator.sort crashes when sorting by \DateTimeImmutable
The explicit check in
https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Iterator/SortViewHelper.php#L218

leads to a fatal-error when passing a `\DateTimeImmutable` field as sort-value.

MK-42:
`Iterator\SortViewHelper` now checks for all forms of `\DateTimeInterface` when trying to deduce a sort value from objects

fixes #1610


NamelessCoder:
Looks reasonable enough - thanks @MK-42 :)

PR #1627, opened 2020-02-26
[FIX] fix Security Viewhelper

samowitsch:
fix for deprecated property loginUser of TypoScriptFrontendController


NamelessCoder:
If you can, please take care of the failing test. The security VH test case uses `stdClass` and will now need to use a mock of `TypoScriptFrontendController` instead, which mocks the `isUserOrGroupSet` method and has it return either true or false, depending on the test.

samowitsch:
> If you can, please take care of the failing test. The security VH test case uses `stdClass` and will now need to use a mock of `TypoScriptFrontendController` instead, which mocks the `isUserOrGroupSet` method and has it return either true or false, depending on the test.

Hi. I looked into the test but i have no clue how to do that.

PR #1646, opened 2020-05-12
Bugfix in SortViewHelper: Exception when providing subject without as

schoeppe:
This is caused by checking the argument as instead of subject...

### DIFF
--- a/Classes/ViewHelpers/Iterator/SortViewHelper.php
+++ b/Classes/ViewHelpers/Iterator/SortViewHelper.php
@@ -109,7 +109,7 @@ class SortViewHelper extends AbstractViewHelper
         \Closure $renderChildrenClosure,
         RenderingContextInterface $renderingContext
     ) {
-        $subject = static::arrayFromArrayOrTraversableOrCSVStatic(!empty($arguments['as']) ? $arguments['subject'] : $renderChildrenClosure());
+        $subject = static::arrayFromArrayOrTraversableOrCSVStatic(!empty($arguments['subject']) ? $arguments['subject'] : $renderChildrenClosure());
         $sorted = null;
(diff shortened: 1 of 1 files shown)


NamelessCoder:
This may be a misunderstanding - the logic of this check is:

* If user provided the `as` argument then `subject` is only possible to find in `$arguments['subject']` since the tag content would be "the code to render with added variable".
* If user did not provide `as` then `$renderChildrenClosure()` will return either the `$arguments['subject']` **or** the tag content (if subject was passed as variable inside the tag, or by inline pass).

When you specify `as` this automatically forces rendering of the tag content with the variable added as for example `{sorted}` if you used `as="sorted"`. The ViewHelper is not going to "leak" the variable, e.g. `<v:iterator.sort subject="{foo}" as="bar" />` does not make `{bar}` available; such an expression is a "no-operation" because the tag content is empty. Same is true for `{foo -> v:iterator.sort(as: 'bar')}` which I suspect is the use case that triggers your error?

In any case: the check and exception is intentionally written that way - are you sure you haven't written your implementation in one of the (mistaken) ways described above?

schoeppe:
I suppose you are right and I used the viewhelper wrong. I had:
```
<f:for each="{v:iterator.sort(subject: menu, order: 'RAND')}" as="menuItem" iteration="iterator">
<!-- loop logic -->
</f:for>
```
I changed this to:
```
<f:for each="{menu -> v:iterator.sort(order: 'RAND')}" as="menuItem" iteration="iterator">
<!-- loop logic -->
</f:for>
```

When I look at the statements intuitively, at least for me, both seem right. But as you pointed out, in the first case subject is taken from `$renderChildrenClosure` and empty.

Thanks you for the detailed explanation and the quick response!

Issue #1648, opened 2020-05-18
Replace $GLOBALS['TSFE']->sys_language_uid in PageService.php


Flimmerkiste:
There is a reference to `$GLOBALS['TSFE']->sys_language_uid`, which was removed in TYPO3 10, in `./Classes/Service/PageService.php` which seems to prevent menus (`<v:menu ...>`) from rendering in my project. Please replace this with language aspect.

breakone:
#1650

NamelessCoder:
@breakone Would you mind adding a `Close: #1648` line to your commit message for PR #1650? That way this current issue will be closed when the PR is merged ;)

PR #1650, opened 2020-05-27
Replace $GLOBALS['TSFE']->sys_language_uid

breakone:
Use LanguageAspect if present

Close: #1648


NamelessCoder:
@breakone Thanks for the patch - changes appear OK but perhaps the tests need to be adjusted on 9.x, it appears we may need to mock the context to test for a specific language? I assume the tests cause the error because the language that gets returned isn't mocked and the default value (presumably, "default") doesn't match what the tests assert as expectation.

breakone:
Close: #1648

breakone:
Hi @NamelessCoder, in case of `LanguageAspect` class is present, the default language id is 0. In the other case `$GLOBALS['TSFE']->sys_language_uid` is null. That means for > 9.x the test returns `then`, for < 9.x `else`.

What do you think about mocking the `TypoScriptFrontendController`, so in all cases the default value is set to 0 and check on `$this->assertEquals('then', $result);`?

`$GLOBALS['TSFE'] = $this->getMockBuilder(TypoScriptFrontendController::class)->disableOriginalConstructor()->getMock();`

Issue #1643, opened 2020-04-30
Update v:page.header.canonical to use setMetaTag() for TYPO3 10


Flimmerkiste:
**Setup**
TYPO3 10.4.1
VHS 6.0.0

Using the `v:page.header.canonical` viewhelper results in an exception: `Call to undefined method TYPO3\CMS\Core\Page\PageRenderer::addMetaTag()`

The `addMetaTag()` method was deprecated in TYPO3 9 and removed in TYPO3 10. The new method to be used is `TYPO3\CMS\Core\Page\PageRenderer::setMetaTag()`

NamelessCoder:
Solved via linked PR.

PR #1651, opened 2020-05-27
[BUGFIX] Replace use of deprecated method.

datamints-off:
PageRenderer::addMetaTag() was deprecated in TYPO3 v9 (and remove in v10), so it's better to use addHeaderData() instead.


Classes/ViewHelpers/Page/Header/LinkViewHelper.php
```diff
@@ -53,6 +53,6 @@ public function render()
         if ('BE' === TYPO3_MODE) {
             return;
         }
-        static::getPageRenderer()->addMetaTag($this->renderTag($this->tagName));
+        $this->getPageRenderer()->addHeaderData($this->renderTag($this->tagName));
```
NamelessCoder, line 56: This seems to be the right solution, at least for cross-compatibility down to 8.7 - but the method should be called statically as it is a static method ;)

datamints-off:
The last commit addresses the issue from #1643

PR #1657, opened 2020-06-04
[BUGFIX] Replaced Deprecated PATH_site Constant, Updated Utility Func…

masonen:
…tion as documented in Typo3core API

- Refactored usage of Deprecated ```PATH_site``` constant in ```AssetService``` and ```AbstractImageViewHelper```
- Changed ```CoreUtitlity::getSitePath()``` to Return a correct Replacement of the PATH_site constant (see: https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/ApiOverview/GlobalValues/Constants/Index.html#path-site)


magicsunday:
I think you should remove the constant() calls too. 

`constant('PATH_site') => CoreUtility::getSitePath()`

masonen:
Youre absolutly right. Will fix today.

Issue #1659, opened 2020-06-08
Security/AllowViewHelper not working with TYPO3 10


freshman20:
Both security ViewHelpers (AllowViewHelper + DenyViewHelper) not working with TYPO3 10.
$GLOBALS['TSFE']->loginUser was marked as deprecated in TYPO3 v9.4 but it's still in -Security/AbstractSecurityViewHelper->getCurrentFrontendUser()

dextar47:
Solution for TYPO3 10 and backward compatible till v8 https://github.com/FluidTYPO3/vhs/pull/1745

PR #1660, opened 2020-06-08
[BUGFIX] fixed Security/AllowViewHelper for TYPO3 v10 (#1659)

freshman20:
Fix for issue 1659
https://github.com/FluidTYPO3/vhs/issues/1659


IchHabRecht:
@freshman20 your patch doesn't cover the compatibility to TYPO3 8.7 which is still marked as a supported version. I guess this pull request can't get merged as it is.

freshman20:
@IchHabRecht
You're right. thanks for the hint.

PR #1661, opened 2020-06-09
Prevent calling deprecated TemplateService->init()

IchHabRecht:


NamelessCoder:
Thanks for the patch @IchHabRecht - do you know if the `TYPO3_version` is still there in 10 LTS and won't be deprecated? If it isn't there or is deprecated we can instead use `ExtensionManagementUtility::getExtensionVersion('core')` to read out the version number.

IchHabRecht:
It got deprecated with 10. Just feeld free to edit the pull request. I just need to get things working over here.

NamelessCoder (approved):

PR #1662, opened 2020-06-11
[BUGFIX] Remove deprecated call to GeneralUtility::unQuoteFilenames

mh0rst:
The deprecated method GeneralUtility::unQuoteFilenames has been removed from TYPO3 v10, which breaks the LinkViewHelper

This PR introduces a copy of aforementioned method. The LinkViewHelper is the only instance using unQuoteFilenames, so I guess the copy should go there, too.


NamelessCoder:
@mh0rst Thanks for the patch - but I think we should instead remove the entire processing block that is entered if $pageUid is not numeric. The `pageUid` argument is specified as `integer` and should not support the typolink syntax with space-separated parts. The parts that are possible to specify this way can also be specified as either direct tag attributes or through `additionalAttributes`.

If you rewrite the patch to drop the added method and remove this condition block then I will happily merge it ;)

mh0rst:
@NamelessCoder you are right, I amended the change set, passing a string to $pageUid now triggers a log warning and returns.

NamelessCoder (approved):

PR #1665, opened 2020-06-22
Method setControllerVendorName is removed

MohsinQK:
setControllerVendorName is removed from TYPO3\CMS\Extbase\Mvc\Web\Request class and also these classes are deprecated and moved to use TYPO3\CMS\Extbase\Mvc\Response , TYPO3\CMS\Extbase\Mvc\Request;


NamelessCoder:
@MohsinQK Thanks for the patch - but we need this ViewHelper to continue supporting TYPO3 8.7 and 9.5 which means we can't just change the class and remove the parameter, as this would break on those versions. Instead we need to have a check if the old class exists and if so, use it. Otherwise, use the replacement class (without the vendorName parameter).

dogawaf:
Hi. 
This quick fix will not work in the end, because in the end the $request will miss the controllerObjectName.

NamelessCoder:
Fixed by https://github.com/FluidTYPO3/vhs/commit/444c2e5f7e2670a1d11867f6180ae9fc07e32317

PR #1671, opened 2020-07-08
[BUGFIX] Respect page language overlay

froehlit:
In `\FluidTYPO3\Vhs\ViewHelpers\Resource\Record\FalViewHelper::getFileReferences`  use `$record['_PAGES_OVERLAY_UID']` to get the correct page uid for the language overlay.

This solution works for me, but I'm not an expert in TYPO3s language handling.

### DIFF
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -37,6 +37,11 @@ class FalViewHelper extends ResourcesFalViewHelper
     protected $field = self::DEFAULT_FIELD;
 
     /**
+     * @var string
+     */
+    protected $localizedIdField = "_PAGES_OVERLAY_UID";
+
+    /**
      * Initialize arguments.
      *
      * @return void
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -62,6 +62,11 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
     protected $escapeOutput = false;
 
     /**
+     * @var string
+     */
+    protected $localizedIdField = "_LOCALIZED_UID";
+
+    /**
      * Constructor
      */
     public function __construct()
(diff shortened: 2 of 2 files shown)


Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -118,6 +118,8 @@ protected function getFileReferences($table, $field, $uidOrRecord)
             $sqlRecordUid = $record['t3ver_oid'];
         } elseif (isset($record['_LOCALIZED_UID'])) {
             $sqlRecordUid = $record['_LOCALIZED_UID'];
+        } elseif (isset($record['_PAGES_OVERLAY_UID'])) {
```
NamelessCoder, line 121: This property name is specific to pages and does not belong in the generic FAL resource ViewHelper. A more appropriate place is in the page-specific FAL resource ViewHelper.

froehlit: If i fixe it only in `v:page.resource.fal` viewhelper, the problem remains, if you call the `v:resource.record.fal` viewhelper with `table="page"` as an argument.

Classes/ViewHelpers/Page/Resources/FalViewHelper.php
```diff
@@ -36,6 +36,11 @@ class FalViewHelper extends ResourcesFalViewHelper
      */
     protected $field = self::DEFAULT_FIELD;
 
+    /**
+     * @var string
+     */
+    protected $localizedIdField = "_PAGES_OVERLAY_UID";
```
NamelessCoder, line 42: I'm not sure this works the way you think. On TYPO3v8 this should be the UID of the `pages_language_overlay` record (correct me if I'm wrong!) - and...

Classes/ViewHelpers/Resource/Record/FalViewHelper.php
```diff
@@ -61,6 +61,11 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
      */
     protected $escapeOutput = false;
 
+    /**
+     * @var string
+     */
+    protected $localizedIdField = "_LOCALIZED_UID";
```
NamelessCoder, line 67: ...on all TYPO3 versions this field should contain the UID of the original record that was translated, not the translated record's UID (which should be in the `uid` property). For non-page records this part should work, assuming that `uid` contains the `uid` of the non-default language record (the methods called with this variable require the original record's UID if I remember it right), but for pages I think the implementation with `_PAGES_OVERLAY_UID` breaks on TYPO3v8 as it doesn't contain the UID of a page, but rather the UID of a `pages_language_overlay` record.

user123801:
@NamelessCoder Any update on the issue? The problem persists in the latest version of the vhs extension, it shows the images of the default language in alternative site languages.

fnagel:
Update: still having issues when using slide and having disabled pages on the rootline.

NamelessCoder:
Fixed by https://github.com/FluidTYPO3/vhs/commit/169a6112fb4625a86c5bcb3b05a7c3b7df361d77

PR #1676, opened 2020-07-25
[BUGFIX] Replaced deprecated GeneralUtility::sysLog calls with new logging API

### LINKED ISSUE #1649: Replace calls to GeneralUtility::sysLog for TYPO3 9.0+
See https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Breaking-82430-ReplacedGeneralUtilitysysLogWithLoggingAPI.html

bink:
Fixes #1649 

- Changed FluidTYPO3\Vhs\Service\AssetService to implement LoggerAwareInterface and call $this->logger->warning() instead of GeneralUtility::sysLog()
- Added static getLogger() function to FluidTYPO3\Vhs\ViewHelpers\Iterator\ExtractViewHelper and used it to call self::getLogger()->warning() instead of GeneralUtility::sysLog()

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -11,6 +11,8 @@ namespace FluidTYPO3\Vhs\Service;
 use FluidTYPO3\Vhs\Asset;
 use FluidTYPO3\Vhs\Utility\CoreUtility;
 use FluidTYPO3\Vhs\ViewHelpers\Asset\AssetInterface;
+use Psr\Log\LoggerInterface;
+use TYPO3\CMS\Core\Log\LogManager;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
@@ -30,7 +32,6 @@ use TYPO3Fluid\Fluid\Core\ViewHelper\TagBuilder;
  */
 class AssetService implements SingletonInterface
 {
-
     /**
      * @var string
      */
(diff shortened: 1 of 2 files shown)


olinox14:
This pull request should be merged, I faced the very same issue, and I solved the bug by manually applying those changes

TrueType:
what about it?

NamelessCoder:
I'd like to merge this but doing so would break compatibility with 8.7. AFAICS it won't be possible to make use of the LoggerAware interface and trait so we would need a different solution for this (one that uses Logger without implementing the trait and interface, and which falls back to GeneralUtility::sysLog in case Logger doesn't exist or TYPO3 version is detected as 8.7.x).

olinox14:
What about a dedicated Logger class that would implement the new 9.5+ interface, and a second class which would simply call the old GeneralUtility::sysLog, the first one would be the default one, but the second would replace it only if typo3 version is inferior to 9?

Could we use XClasses to implement this? Something like:

```
if ((int) explode('.', TYPO3_version)[0] < 9 ){
    $GLOBALS['TYPO3_CONF_VARS']['SYS']['Objects'][FluidTYPO3\Vhs\Utility\LoggerUtility::class] = [
        'className' => FluidTYPO3\Vhs\Utility\Retrocompatibility\LoggerUtility::class
    ];
}
```

NamelessCoder:
@olinox14 Since there are only two instances of GeneralUtility::sysLog being used, a Utility is overkill. I'd rather see a condition with a version check in those two places, where LogManager is used to retrieve a Logger instance if TYPO3 version is >= 9.5. This also means that the interface and trait should not be used because this will cause class parsing failures on versions where these are not available.

NamelessCoder (approved):

PR #1678, opened 2020-08-07
[BUGFIX] prevent double encoded sourceset urls

### LINKED ISSUE #1674: Double encoded image urls when using SourceSetViewHelperTrait
Related to #693, #1364 

I have a FAL file with special chars as input for `<v:media.image src="{image.id}"` and srcset configuration.

When \FluidTYPO3\Vhs\Traits\SourceSetViewHelperTrait::addSourceSet runs, it already rawurlencodes the url when it calls static::preprocessSourceUri, then preprocessSourceUri runs rawurlencode again.

Result: "Lösungen" becomes "L%25C3%25B6sungen" instead of "L%C3%B6sungen".
I guess the encoding in addSourceSet could be dropped

akiessling:
Fixes #1674


luzat (approved):
This is similar (albeit updated) to my previously proposed solution and I don't see this change having the ability to break anything, but to fix the double-encoding for sure. @NamelessCoder Would it possible to review and merge this? The issue now exists for 4+ years and does have a simple fix.

Issue #1680, opened 2020-08-14
PageService->getRootline ignores $pageUid since TYPO3 v10 (+Solution)


dev-rke:
Hi there,

thanks for this extension, it saved me a lot of time in recent years. :-)
Since vhs 6 i found out, that there is a problem resolving rootlines since TYPO3 v10.

### Environment
vhs v6.0.3
TYPO3 v10.4.6

### Current Behaviour
When calling the `PageService->getRootline()` method with the `$pageUid` argument, the full rootline gets returned. Always.
The `$pageUid` parameter is ignored.
This affects all situations where the rootline is used, e.g. `v:menu()`, `v:page.rootline()` or the `SlideViewHelperTrait`.

### Expected Behaviour
The rootline should be filtered to begin at the given pageUid and should not return all pages. 

### Issue analysis
Lets have a look at the current implementation:
```php
if (method_exists($pageRepository, 'getRootLine')) {
    if (true === (boolean)$disableGroupAccessCheck) {
        $pageRepository->where_groupAccess = '';
    }
    $rootline = $pageRepository->getRootLine($pageUid);
} elseif (isset($GLOBALS['TSFE'])) {
    $rootline = (array)$GLOBALS['TSFE']->rootLine;
} else {
    $rootline = [];
}
```
The code will try to resolve rootline from `PageRepository` by default and filters for pageUid.
When the `PageRepository` does not exist (since TYPO3 v10), it will resolve rootline from TSFE, but does NOT filter by the given `$pageUid`. Therefore a different (and wrong) behaviour is created.
This issue will not rise up on TYPO3 v9, as the `PageRepository->getRootline()` method still exists and will be called by default.

### Solution
My recommendation is to rewrite the code using the new `RootlineUtility`, which is available since TYPO3 v9. 
Futhermore i switched default behaviour to the `RootlineUtility`, as the logic around `PageRepository` should be removed as soon as TYPO3 8 will not be supported anymore by vhs:
```php
if(class_exists(RootlineUtility::class)) {
    $rootline = (new RootlineUtility($pageUid))->get();
} elseif (method_exists($pageRepository, 'getRootLine')) {
    if (true === (boolean) $disableGroupAccessCheck) {
        $pageRepository->where_groupAccess = '';
    }
    $rootline = $pageRepository->getRootLine($pageUid);
} else {
    $rootline = [];
}
```

I will provide a pull request on this.

NamelessCoder:
Merged - thanks!

[...]

PR #1681, opened 2020-08-14
[BUGFIX] #1680 fixed rootline to filter correct by pageUid

dev-rke:
also use RootlineUtility by default


NamelessCoder (approved):

NamelessCoder:
Thanks for the patch, @dev-rke!

PR #1691, opened 2020-10-20
Support core FileReference in image size viewhelpers

mbrodala:


NamelessCoder (approved):

NamelessCoder:
Thanks Mathias, good catch!

PR #1692, opened 2020-11-20
PHP 7.3.0 deprecated behavior

ghanshyamgohel:
If needle is not a string, it is converted to an integer and applied as the ordinal value of a character. This behavior is deprecated as of PHP 7.3.0, and relying on it is highly discouraged. Depending on the intended behavior, the needle should either be explicitly cast to string, or an explicit call to chr() should be performed.


dmitryd:
I am wondering when it can be something else. It is supposed to be a string always...

NamelessCoder:
> I am wondering when it can be something else. It is supposed to be a string always...

Fluid detects numeric values and creates a NumericNode which can contain integer/float, and if passing a variable reference it may also resolve to null or even objects.

@ghanshyamgohel Pre-approved but would you mind switching it to cast `(string) $arg` instead of strval and applying the same casting to the haystack as well?

ghanshyamgohel:
@NamelessCoder Thank you for Pre-approved. I have changed `strval` to `(string)`

Issue #1697, opened 2021-01-06
v:render.request broken on TYPO3v10


mbrodala:
The `RequestViewHelper` calls `Request::setControllerVendorName()` which has been [dropped with TYPO3v10](https://github.com/TYPO3/TYPO3.CMS/commit/be8d42d6ed09cf400856b80c26021387e85320e7):

> Call to undefined method TYPO3\CMS\Extbase\Mvc\Web\Request::setControllerVendorName()

Due to this the viewhelper is currently not usable on TYPO3v10.

develth:
Any News on this? 
Patch is working  (thanks @mbrodala), but would be nice to have this in the repo

typecat:
Following patch worked for me: remove `$request->setControllerVendorName($vendorName);` and to set a controller alias via `$request->setControllerAliasToClassNameMapping.`

NamelessCoder:
Solved by https://github.com/FluidTYPO3/vhs/pull/1707

PR #1698, opened 2021-01-06
[BUGFIX] v:render.request broken on TYPO3v10

monosize:
check TYPO3 version
fix for #1697


monosize:
Close this pull request, cause it did not work on typo3 10

PR #1700, opened 2021-01-21
[TASK] Add Service.yaml to expose classes as services for DI

Konafets:
When migrating from `$objectManager->get()` to constructor dependency injection I got this error:

> Cannot autowire service "Vendor\ExtensionKey\ViewHelpers\MenuViewHelper": argument "$pageService" of method "FluidTYPO3\Vhs\ViewHelpers\Menu\AbstractMenuViewHelper::injectPageService()" references class "FluidTYPO3\Vhs\Service\PageService" but no such service exists.

To be able to use constructor DI, the service needs to be exposed through an Service.yaml file. This PR adds this file.

### DIFF
--- a/Configuration/Services.yaml
+++ b/Configuration/Services.yaml
@@ -1,0 +1,8 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+
+  FluidTYPO3\Vhs\:
+    resource: '../Classes/*'


NamelessCoder (changes requested):
This is a bit too broad - not every class is a service you should be able to DI and autowiring every ViewHelper class is going to add a lot of unnecessary load and size to the DI manifest. I would suggest limiting the autowiring to Classes/Service and perhaps include Classes/View.

NamelessCoder:
VHS now has an appropriate set of service definitions, public=true is set for injectable classes and ViewHelpers.

PR #1707, opened 2021-02-19
[BUGFIX] v:render.request broken on TYPO3v10

monosize:
insert modified loadDefaultValues based on \TYPO3\CMS\Extbase\Mvc\Web\RequestBuilder::loadDefaultValues
fix for #1697

### DIFF
--- a/Classes/ViewHelpers/Render/RequestViewHelper.php
+++ b/Classes/ViewHelpers/Render/RequestViewHelper.php
@@ -87,17 +87,6 @@ class RequestViewHelper extends AbstractRenderViewHelper
         );
 
         $temporaryContentObject = GeneralUtility::makeInstance(ContentObjectRenderer::class);
-        /** @var Request $request */
-        $request = $objectManager->get(static::$requestType);
-        $request->setControllerActionName($action);
-        $request->setControllerName($controller);
-        $request->setPluginName($pluginName);
-        $request->setControllerExtensionName($extensionName);
-        if (!empty($requestArguments)) {
-            $request->setArguments($requestArguments);
-        }
-        $request->setControllerVendorName($vendorName);
-
         try {
             /** @var ResponseInterface $response */
             $response = $objectManager->get(static::$responseType);
@@ -109,6 +98,21 @@ class RequestViewHelper extends AbstractRenderViewHelper
                     $pluginName
                 )
             );
+
+            if (version_compare(TYPO3_version, 10.0, '<')) {
+                /** @var Request $request */
+                $request = $objectManager->get(static::$requestType);
+                $request->setControllerActionName($action);
+                $request->setControllerName($controller);
+                $request->setPluginName($pluginName);
+                $request->setControllerExtensionName($extensionName);
+                if (!empty($requestArguments)) {
+                    $request->setArguments($requestArguments);
+                }
+                $request->setControllerVendorName($vendorName);
+            } else {
+                $request = self::loadDefaultValues($extensionName, $pluginName, $action, $requestArguments);
+            }
             static::getDispatcher()->dispatch($request, $response);
             $configurationManager->setContentObject($contentObjectBackup);
             if (true === isset($configurationBackup)) {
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Render/RequestViewHelper.php
```diff
@@ -149,4 +153,46 @@ protected static function getConfigurationManager()
...
+            $controllerAliasToClassMapping[$controllerConfiguration['alias']] = $controllerConfiguration['className'];
+            $controllerClassToAliasMapping[$controllerConfiguration['className']] = $controllerConfiguration['alias'];
+        }
+
+        /** @var \TYPO3\CMS\Extbase\Mvc\Web\Request $request */
+        $request = static::getObjectManager()->get(Request::class);
+        $request->setPluginName($pluginName);
+        $request->setControllerExtensionName($extensionName);
+        $request->setControllerAliasToClassNameMapping($controllerAliasToClassMapping);
+        $request->setControllerName($controllerClassToAliasMapping[$controllerClassName]);
```
RozbehSharahi, line 186: I might be wrong, but this looks wrong to me.

Why would you iterate through 'controllerConfigurations' and then take the last controller in that configuration (see forEach) as the controllerName?

Shouldn't you rather pass the controllerName given by arguments to this method and use it instead?

Classes/ViewHelpers/Render/RequestViewHelper.php
```diff
@@ -149,4 +153,46 @@ protected static function getConfigurationManager()
...
+     * @param $pluginName
+     * @param $actionName
+     * @param $parameters
+     * @throws MvcException\InvalidActionNameException
+     * @throws MvcException\InvalidArgumentNameException
+     * @throws MvcException\InvalidControllerNameException
+     * @throws MvcException\InvalidExtensionNameException
+     * @see \TYPO3\CMS\Extbase\Core\Bootstrap::initializeConfiguration
+     */
+    protected static function loadDefaultValues($extensionName, $pluginName, $actionName, $parameters)
```
develth, line 168: There should be a $controllerName Parameter, so you can use it at line 186

Classes/ViewHelpers/Render/RequestViewHelper.php
develth, line 186: When using a $controllerName parameter it would be enough to use

`$request->setControllerName($controllerName);`

Classes/ViewHelpers/Render/RequestViewHelper.php
```diff
@@ -149,4 +153,46 @@ protected static function getConfigurationManager()
...
+    {
+        $configurationManager = static::getConfigurationManager();
+        $configuration = $configurationManager->getConfiguration(ConfigurationManagerInterface::CONFIGURATION_TYPE_FRAMEWORK, $extensionName, $pluginName);
+        $configuration['extensionName'] = $extensionName;
+        $configuration['pluginName'] = $pluginName;
+
+        $controllerAliasToClassMapping = [];
+        foreach ($configuration['controllerConfiguration'] as $controllerClassName => $controllerConfiguration) {
+            $controllerAliasToClassMapping[$controllerConfiguration['alias']] = $controllerConfiguration['className'];
+            $controllerClassToAliasMapping[$controllerConfiguration['className']] = $controllerConfiguration['alias'];
```
develth, line 178: IMHO this is not needed, when using $controllerName parameter

Classes/ViewHelpers/Render/RequestViewHelper.php
```diff
@@ -109,6 +98,21 @@ public static function renderStatic(
...
+                $request->setControllerActionName($action);
+                $request->setControllerName($controller);
+                $request->setPluginName($pluginName);
+                $request->setControllerExtensionName($extensionName);
+                if (!empty($requestArguments)) {
+                    $request->setArguments($requestArguments);
+                }
+                $request->setControllerVendorName($vendorName);
+            } else {
+                $request = self::loadDefaultValues($extensionName, $pluginName, $action, $requestArguments);
```
develth, line 114: Add e.g. after `$extensionName` the `$controller` parameter

develth (changes requested):
Added some feedback to use the available $controller argument as parameter

Issue #1713, opened 2021-03-29
BUG: sort viewhelper "subject"/"as handling


helsner:
I got the issue that the sortviewhelper returns null when using the "subject" argument instead of the children.
`$subject = static::arrayFromArrayOrTraversableOrCSVStatic(!empty($arguments['as']) ? $arguments['subject'] : $renderChildrenClosure());`
in 
https://github.com/FluidTYPO3/vhs/blob/development/Classes/ViewHelpers/Iterator/SortViewHelper.php#L112
returns the children closure even if subject is set due to the necesseary "as".
But adding "as" then fails later on while processing the name of the variable when calling 
`return static::renderChildrenWithVariableOrReturnInputStatic(...` at the end of the function

The current workaround is to only use the children closure handling.

NamelessCoder:
Very true, this looks like a leftover from a time when the render children closure did not take the "content argument" into account. I would appreciate a pull request that changes the ternary condition to simply calling `$renderChildrenClosure()` to get the input argument for the array resolving function!

PR #1715, opened 2021-03-29
[BUGFIX] Remove subject and enforce childrenClosure rendering in sort vh

helsner:
Refs #1713


NamelessCoder (approved):

NamelessCoder (changes requested):
May need to revisit this one, I initially overlooked that the ViewHelper does not use the content-argument trait which would cause $renderChildrenClosure() to return either the argument or the (inline) content. Not sure if the solution is to implement that trait, or reproduce the logic it provides.

PR #1717, opened 2021-04-01
Enable stable releases from TYPO3-11-branch with composer

buepro:
When upgrading an extension for TYPO3-11 it is desirable to work with as much stable releases as possible. This pull request would allow to use the dev-branch from vhs together with stable TYPO3 releases from the latest branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,9 +60,6 @@
           "cms-package-dir": "{$vendor-dir}/typo3/cms",
           "extension-key": "vhs",
           "web-dir": "build/web"
-        },
-        "branch-alias": {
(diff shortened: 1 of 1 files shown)


dmitryd:
@NamelessCoder Can this be merged? It is really hard to use EXT:vhs with v11 now because of restrictions in `composer.json`. This will let everybody prepare to v11 LTS in advance.

Thanks! 🙂

NamelessCoder (approved):

NamelessCoder:
@dmitryd Sure, here you go! And thanks @buepro :)

PR #1724, opened 2021-06-07
better English

franzholz:


NamelessCoder:
I've made a couple of changes; some of the suggested changes were based on misreading. For the record, the by far more common of "inside" or "inside of" is actually "inside" (this is the right form when used as preposition). The form "inside of" is actually a noun (the inside of X). In our case, the meaning is a preposition.

NamelessCoder:
Processed and merged locally

PR #1727, opened 2021-07-06
Fixed possible permission problem with cache:flush

schoeppe:
The touch command only succeeds if the process user is the file owner.
This must not be true in some deployment scenarios: The deployment
user is allowed to write (and delete) such cache files but must not be
the owner.

The workaround is to recreate the cache file (in an atomic operation)
with the same contents and touch the newly created file.

See FluidTYPO3#1509.


NamelessCoder (approved):

NamelessCoder:
Thanks for the solution, @schoeppe :)

PR #1732, opened 2021-08-17
[BUGFIX] Allow new extension configuration in ExtensionConfigurationViewHelper

Lagerregal:
Since TYPO3 is using '$GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS'][$extensionKey]'
instead of the serialized '$GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'][$extensionKey]'
the ExtensionConfigurationViewHelper should support this new configuration structure.


NamelessCoder:
Thanks for the patch @Lagerregal!

PR #1733, opened 2021-09-08
Add v:context.get() support for 11.4

glucka:


NamelessCoder (changes requested):
We need a version or class exists check here to use Environment::getContext only when it is available.

NamelessCoder:
Thanks for taking care, @glucka - I processed this manually and added fallbacks for TYPO3 versions without the Environment class, in https://github.com/FluidTYPO3/vhs/commit/b393c68653722f4267cf08f6d507b8eef9cd19fc

PR #1738, opened 2021-10-10
[TASK] Add new page repository alias

tobsnti:
Because PageRepository has been moved with TYPO3 11 the class alias is created based on class existence.

### DIFF
--- a/Classes/Service/PageService.php
+++ b/Classes/Service/PageService.php
@@ -14,7 +14,11 @@ use TYPO3\CMS\Core\Context\LanguageAspect;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Utility\RootlineUtility;
-use TYPO3\CMS\Frontend\Page\PageRepository;
+if (class_exists(\TYPO3\CMS\Core\Domain\Repository\PageRepository::class)) {
+    class_alias('TYPO3\CMS\Core\Domain\Repository\PageRepository', __NAMESPACE__ . '\PageRepository');
+} else {
+    class_alias('TYPO3\CMS\Frontend\Page\PageRepository', __NAMESPACE__ . '\PageRepository');
+}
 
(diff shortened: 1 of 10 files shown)


NamelessCoder:
Sorry for the delay here, @tobsnti - I've been very busy elsewhere.

Re: this PR, I would very much prefer to have a Factory that gives us the right instance based on presence of one class or the other. Something like this (freestyle):

```php
class InstanceFactory {
    /**
     * @return \TYPO3\CMS\Frontend\Page\PageRepository|\TYPO3\CMS\Core\Domain\Repository\PageRepository
     */
    public function getPageRepository()
    {
        if (class_exists(\TYPO3\CMS\Frontend\Page\PageRepository::class)) {
            return GeneralUtility::makeInstance(\TYPO3\CMS\Frontend\Page\PageRepository::class);
        }
        return GeneralUtility::makeInstance(\TYPO3\CMS\Core\Domain\Repository\PageRepository::class);
    }
}
```

And then use this factory in all places where we need a PageRepository.

Even though at first this `InstanceFactory` would only have a single method to get the PageRepository, I'm convinced it won't be the last time we run into a requirement of this type so imho it makes a lot of sense to have a factory like this, that would allow us to "hide away" the actual class name that's returned as well as any differences in things like required constructor arguments or details about how the instance is constructed (makeInstance or ObjectManager or DI container fetch, for example). In short, I'd prefer if we handle this with "real" code instead of using metadata files or aliasing. This also has the benefit of working the exact same way in composer/non-composer and unit test context. It requires more refactoring in places where we use the PageRepository but I think it's a better solution in the long run.

helsner:
I guess this will not properly work in the long run, at least IMHO
(https://github.com/FluidTYPO3/vhs/blob/04d7ac203d7726738b280ee2f2eec2ae24c73b48/Classes/ViewHelpers/Page/Resources/FalViewHelper.php#L78)
Here ->init() is called which is protected in v11, thus not callable.
So this would need a workaround as well.
I will llok after the holidays

PR #1745, opened 2021-11-23
[Bugfix] fixed Security/AllowViewHelper for TYPO3 v10 with backward compatibility

dextar47:
Security/Allow viewhelper is not working in TYPO3 10. This solution fixes for TYPO3 10 along with TYPO3 8 compatibility as well.


dextar47:
will fix https://github.com/FluidTYPO3/vhs/issues/1659

NamelessCoder (approved):

NamelessCoder (changes requested):
Needs a fix for the unit test - I suggest refactoring this to add `protected function isFrontendUserLoggedIn` which can be mocked in the failing unit test to force a true/false return value to match the function call expectation.

Issue #1742, opened 2021-11-02
TYPO3 v11 compatibility


mhirdes:
Is there any plan to have a extension version ready for TYPO3 v11 LTS? Do you need help?

coffeecode-de:
> Is there any plan to have a extension version ready for TYPO3 v11 LTS? Do you need help?

I can help too.

NamelessCoder:
Thanks for the offer to help! The most important issue is https://github.com/FluidTYPO3/vhs/pull/1738 but if you know about anything else that also needs to be updated, then pull requests for those things would be *very* appreciated!

velletti:
mabe rising the required version from ^11.1@dev to ^11 instead of : 
   
    "typo3/cms-core": "^8.7 || ^9 || ^10 || ^11.1@dev || dev-master",

robsonrobi:
Yes, a new version would be great. I have vhs locally added and https://github.com/FluidTYPO3/vhs/pull/1738 ist the only necessary change...

sneopu:
A v11 release would be great. Thanks for your efforts.

msollmann:
I would be happy too about a v11 version :-)

tomashavner:
Our upgrade to Typo3 v.11 depends on a new version of vhs. Would really appreciate it.

h-e-l-l-o-w-o-r-l-d:
+1 for an official v11 release. Don't like to work with dev branches.

wayheyha:
If you guys need some extra coffee i could donate some!! :-) 
This is finaly the only dependency left for me to update some systems. 

You guys are awesome!

phifa:
can clients help to speed this up by sponsoring for example?

mrk131324:
The biggest hurdle is to get a full list. The most common issue is an undefined array key in `$this->arguments` (a PHP8 issue). It happens so often, that we need the test suite to find them all.

phifa:
is it safe to use the dev version in prod, when only using the `v:uri.image` viewhelper? I just need the `canvas` functionality.

fgerards:
can we support your efforts to make vhs v11 compatible ?

NamelessCoder:
It took a while but it's finally here - VHS 6.1 is now released, with v11 support.

PR #1749, opened 2021-12-26
[TASK] Provide v8-v11 compatible PageRepository handling

helsner:
Relates #1742 #1738

### DIFF
--- a/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
+++ b/Classes/ViewHelpers/Page/Resources/FalViewHelper.php
@@ -8,6 +8,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Page\Resources;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use FluidTYPO3\Vhs\Factory\InstanceFactory;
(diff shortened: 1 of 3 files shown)


Classes/ViewHelpers/Page/Resources/FalViewHelper.php
```diff
@@ -74,8 +75,10 @@ public function getRecord($id)
             if (TYPO3_MODE === 'FE') {
                 $pageRepository = $GLOBALS['TSFE']->sys_page;
             } else {
-                $pageRepository = GeneralUtility::makeInstance(PageRepository::class);
-                $pageRepository->init(false);
+                $pageRepository = InstanceFactory::getPageRepository();
+                if (class_exists(\TYPO3\CMS\Frontend\Page\PageRepository::class)) {
+                    $pageRepository->init(false);
```
cumuru, line 80: init() is now called implicitely in the constructor (see https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5/Deprecation-86338-ChangeVisibilityOfPageRepository-init.html). Wouldn‘t it make sense to let the factory take care of the init call for legacy code?

PR #1754, opened 2022-02-02
apply fix for v:resource.record.fal: Respect page language #1670

### LINKED ISSUE #1670: v:resource.record.fal: Respect page language
Setup
TYPO3 9.5.19
VHS 6.0.3

After the commit [f70baf9](https://github.com/FluidTYPO3/vhs/commit/f70baf9eede41f7f4a4287b7af66130767127a1a), the `v:page.resources.fal` viewhelper only provides the fal resources of the main language.
I think the problem cause by the methode `getFileReferences` in class `\FluidTYPO3\Vhs\ViewHelpers\Resource\Record\FalViewHelper`. It doesn't use the language overlay of the page, to get the correct files from file reprository.

hacksch:
[BUGFIX] v:resource.record.fal: Respect page language

Get record overlay

Close #1670


NamelessCoder (changes requested):
This will need a rebase and possibly conflict resolving; a recent commit changed how the PageRepository is used - the good news is that it should make this change much smaller, too.

PR #1761, opened 2022-04-11
 Update AssetService.php for PHP 8

warki:
Bugfix for PHP8 while updating a content element


balti02:
how can i Install php 7/8 Windows 10

PR #1763, opened 2022-04-27
TYPO3 V10 support for ExtensionConfigurationViewHelper

MohsinQK:
as in TYPO3 V10 now extConf is moved to $GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS'] so now it will check for the configuration on this path if not available then check in the old path.


NamelessCoder:
Hi @MohsinQK - thanks for the PR! As far as I'm aware, this was already taken care of in https://github.com/FluidTYPO3/vhs/pull/1732 (by adding the new location as first candidate for lookup) which means this PR shouldn't be necessary. Is there a problem with that solution or did you perhaps create this PR based on an earlier state which didn't include the linked fix?

MohsinQK:
@NamelessCoder Yes that's what I did too. maybe that Pull request is not merged and released on the latest version so that's why I have created another PR, I will close this PR.

PR #1762, opened 2022-04-12
[DOCS] Align with new TYPO3 documentation standards

alexander-nitsche:
Hi Claus, hi Björn, hi Cedric, hi VHS team,

this PR is a suggestion to align your extension documentation with the recently introduced [TYPO3 documentation standards](https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/FileStructure.html) and similar extensions.

The focus is on having one central documentation at Documentation/ and reduce the README to an abstract, badges and further links to all important aspects of the extensions (TER, docs, VCS). 

What is missing: The whole documentation should be rendered and deployed on docs.typo3.org - in perspective. It is currently deployed to the custom page https://viewhelpers.fluidtypo3.org/fluidtypo3/vhs/ and thus not found by the new global search at docs.typo3.org. If you need assistance on how to migrate, i would be happy to support you.

Greetings, Alex

Related: https://github.com/TYPO3-Documentation/T3DocTeam/issues/182
Related: https://github.com/TYPO3-Documentation/T3DocTeam/issues/185

PR #1769, opened 2022-05-08
[TASK] Transform into full documentation

alexander-nitsche:
In the [first step](https://github.com/FluidTYPO3/vhs/pull/1762) the README.md was aligned to the documentation standards and published at docs.typo3.org. In this second step the documentation is transformed into a full documentation, making use of all advantages of the TYPO3 Sphinx theme and being searchable by the global search on docs.typo3.org.

What remains: Rendering the view helper documentation automatically into the Documentation/ViewHelpers/ folder.


alexander-nitsche:
How it looks:
Screen Shot 2022-05-08 at 12 21 13

NamelessCoder:
Thanks for your work @alexander-nitsche!

Is there a way to avoid converting the changelog MD files? These are both automatically generated and have permalinks in various places (which are outside our control). I would even be OK with the changelog item just being a link to the GitHub listing of the MD files.

alexander-nitsche:
No problem. The TYPO3 Sphinx theme support Markdown in a reduced way, which is almost fine for the changelogs (only problem are the links to the diff view of the patch). It is still important to have the changelog integrated into the full documentation so that i added a note about rendering flaws and that the user could navigate to the GitHub view in doubt (amended in commit "[[DOCS] Integrate changelog into full documentation](https://github.com/FluidTYPO3/vhs/pull/1769/commits/6dd284ee3ed07408d4c55e06f48bb781cfcc045c)").

NamelessCoder:
Still a minor problem - the changelog index is also automatically generated as MD - and it looks like the files in Documentation/Classes were dropped without conversion?

NamelessCoder (approved):

PR #1777, opened 2022-05-12
Update AbstractContentViewHelper.php

Werghemmi:
resolve PHP 8.1 error: undefined array key 'tt_content:xx'


GuidoJansenPI:
I had the same error and this change fixed it for me

PR #1789, opened 2022-06-20
[BUGFIX] Check flagfile existence before rendering

### LINKED ISSUE #1788: getLanguageFlag() assumes filename corresponding to ISO code.
When the flag file is rendered in `getLanguageFlag($iso, $label)`, it assumes that a flag file for the corresponding two letter iso code (iso-639-1) exists. This is sometimes not the case, a very prominent example is English. With the correct iso-639-1 code `en` in the site configuration, `getLanguageFlag($iso, $label)` tries to render the image EN.png (or another file extension if the user provided a different ImageType).
If no further path has been configured, the path to the core flag icons is used. This triggers a warning on every rendering of the Language Menu because no `EN.png` icon exists. There is only a `GB.png` icon.
To prevent the warning from being triggered, I propose a check, if the file exists and if not, return an empty String.
I will submit a PR for that.

What also bothers me, is that the flag is rendered in any case. I think that when the user provides the layout `name`, that should not be required? Or am I overlooking something there?
If you are interested in a PR for that, too, please comment in this Issue. I will then start a new issue and PR.

cschnell:
Fixes #1788


NamelessCoder:
Thanks @cschnell - I've processed this one manually, you will need to re-checkout the development branch on your local/fork, or fork/clone the repository again. For future patches I recommend creating a new branch for your commit so you won't have problems with sync of your development branch.

PR #1801, opened 2022-08-15
[BUGFIX] fix problems with PHP 8 and TYPO3 11.5

hannesbochmann:


NamelessCoder (approved):

NamelessCoder:
@hannesbochmann Thanks - please pull from origin w/ rebase and re-push the branch. For future patches I highly recommend creating a separate branch so you don't commit to your `development` branch, and to enable edits from maintainers in pull request settings.

hannesbochmann:
@NamelessCoder Done.

NamelessCoder:
I've processed the pull request manually. You will now need to recreate or force-push your fork's `development` branch to bring your fork up to date.

Small note: The change in `\FluidTYPO3\Vhs\ViewHelpers\Render\AbstractRenderViewHelper::renderView` to null-coalesce `graceful` should never trigger - every caller of this method is from a subclass of the abstract class and all subclasses use `parent::initializeArguments` which defines the `graceful` argument with a default value. As such, the value should always exist. If this caused you problems because you created a third-party subclass of this abstract class and called `renderView` I would highly recommend making that subclass call `parent::initializeArguments` in its `initializeArguments` method so it can utilise the graceful handling.

PR #1807, opened 2022-10-05
[BUGFIX] PHP 8 warning fix if $record is empty in getResources func

dextar47:
I upgraded my project's PHP version to PHP 8.1 and got this PHP warning in the backend layout module. The issue was that $record in getResources function was empty (ViewHelpers/Resource/Record/FalViewHelper.php) and then there was no $record[$this->idField] was trying to fetch uid key's value in $record array (without isset() in else part, Line no. 147).

### DIFF
--- a/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
+++ b/Classes/ViewHelpers/Resource/Record/FalViewHelper.php
@@ -133,7 +133,7 @@ class FalViewHelper extends AbstractRecordResourceViewHelper
      */
     public function getResources($record)
     {
-        if (!is_array($record)) {
+        if (!is_array($record) || empty($record)) {
             return [];
         }
         if (!empty($GLOBALS['TSFE']->sys_page)) {


NamelessCoder (approved):

dextar47:
@NamelessCoder thanks for approving the PR. But it shows all checks have failed. And it failed with this message "typo3/class-alias-loader contains a Composer plugin which is blocked by your allow-plugins config". I don't think so that it is due to this PR change. Is it intentional or know issue?

NamelessCoder:
Hi @dextar47 - it's unrelated to your change, the cause is a composer change that requires every composer plugin to be whitelisted before install can happen. I think the problem is already fixed in VHS upstream so if you could either rebase and push your patch branch again, or grant me access to do it for you (settings in PR to allow maintainers of upstream to make changes in your PR branch) then that should allow the build to run.

dextar47:
@NamelessCoder its done 
image

Issue #1816, opened 2022-12-06
PHP Warning: Undefined array key "tt_content:xxxx"


sk-foresite:
When using the `v:content.render` ViewHelper I get the following warning:

```
PHP Warning: Undefined array key "tt_content:3078" in /home/<REDACTED>/www/<REDACTED>/public/typo3conf/ext/vhs/Classes/ViewHelpers/Content/AbstractContentViewHelper.php line 239
```

Using:
TYPO3 11.5.19
PHP 8.1
Using branch development (commit 7752545e2f4e1d651a22bd0b523557eac95ebfff at the time of posting).

It seems to be similiar to issue #1779.

I managed to fix the issues locally by checking that the array key exists:
```php
if (false === empty($parent)) {
    if(array_key_exists($parent, $GLOBALS['TSFE']->recordRegister)) {
        ++$GLOBALS['TSFE']->recordRegister[$parent];
    }
}
$html = $GLOBALS['TSFE']->cObj->cObjGetSingle('RECORDS', $conf);

$GLOBALS['TSFE']->currentRecord = $parent;
if (false === empty($parent)) {
    if(array_key_exists($parent, $GLOBALS['TSFE']->recordRegister)) {
        --$GLOBALS['TSFE']->recordRegister[$parent];
    }
}
```

lucmuller:
I have the same behaviour behaviour using :
- TYPO3 11.5.21
- Php 8.1.7
- VHS 6.1.2

DanYouCan:
With TYPO3 V11.5.21 / PHP8.1.13 / VHS 6.1.3 got same php error:
Core: Error handler (FE): PHP Warning: Undefined array key "tt_content:xyz" in /public/typo3conf/ext/vhs/Classes/ViewHelpers/Content/AbstractContentViewHelper.php line 239

When I use @sk-foresite's workaround, this error disappears.

hbrz:
I've got the same error with TYPO3 11.5.21 / PHP 8.1.3 / VHS 6.1.3
I use VHS in my own extension.

For me, it helped to add my plugin to non-cachable actions:

```
// non-cacheable actions
    array(
        'ControllerName' => 'list, show, create, update',
        
    )
```

Otherwise, the "renderRecord()" get called twice and throws this error.

helsner:
I got this issue as well.
the PR seems good to me as far as fixing the PHP exception that blocks the whole page..

NamelessCoder:
Solved in `dev-development`.

PR #1818, opened 2022-12-06
Fix undefined array key warning

sk-foresite:
Should fix: #1816 

I'm not used to the TSFE registers, so there might be a better way to solve this.


helsner (approved):

NamelessCoder:
Thanks for the pull request - issue is solved via another commit.

PR #1848, opened 2023-03-05
[BUGFIX] Fix notice

georgringer:
The key `hasSubPages` might not be set which leads to an exception


NamelessCoder:
Has been solved in `dev-development` since c78039646dc1dd43a4c070bb060a17c2102be777

Issue #1846, opened 2023-02-21
v:content.render not compatible with PHP 8


darthnorman:
Exception from line 239 after upgrading to PHP 8.

PHP Warning: Undefined array key "tt_content:38796" in /srv/www/public/typo3conf/ext/vhs/Classes/ViewHelpers/Content/AbstractContentViewHelper.php line 239

TYPO3 11.5.24
VHS 6.1.3
PHP 8.1.15

RKlingler:
I'm getting the same same exception from line 239 after upgrading to PHP 8.

> PHP Warning: Undefined array key "tt_content:38796" in /srv/www/public/typo3conf/ext/vhs/Classes/ViewHelpers/Content/AbstractContentViewHelper.php line 239

TYPO3 11.5.24
VHS 6.1.3
PHP 8.1.15

In #1787 someone already mentioned a potential fix for the issue in this particular case.

In PHP 8 the attempt to read from an array with an undefined array key was converted from a notice into a warning. Code should generally be tested for this, as there might be more lines that throw warnings.

NamelessCoder:
Fixed in `dev-development`.

PR #1850, opened 2023-03-14
Fixes AbstractContentViewHelper under PHP 8

DavidR93:
Fixes AbstractContentViewHelper under PHP 8: #1846 

I also added a change to prevent an exception under PHP 8 if the argument "alternative" of OrViewHelper is null. If the $alternative variable is null, then null is passed as a string type parameter ($haystack) in the strpos() function, which results in an exception.


NamelessCoder:
Thanks for the patch @DavidR93 - I've replaced the null-coalesce to empty string with an early return, since an alternate value of `null` needs to be supported (for example, when chaining multiple calls to `v:or` to check multiple variables and use the first non-null value).

For future patches I would recommend creating a new branch in your forked repository and commit fixes to there, then create the PR based on that branch. This saves you from having to hard reset or re-checkout your `development` branch in case commits need to be squashed (which is the default merge strategy in VHS).

NamelessCoder (approved):

PR #1855, opened 2023-04-10
[BUGFIX] Fix Problems with TYPO3 v12.x and cached pages

Neowar:
If a page is cached, the determination of the TypoScript setup fails. A RuntimeException with the code 1666513645 is thrown (\TYPO3\CMS\Core\TypoScript\FrontendTypoScript()->getSetupArray()). This happens in the context of the middleware AssetInclusion.

To prevent this, the exception must be caught in the AssetService in the getSettings() function.

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -136,7 +136,7 @@ class AssetService implements SingletonInterface
                     $allTypoScript['plugin.']['tx_vhs.']['settings.'] ?? []
                 );
             } catch (\RuntimeException $e) {
-                // If RuntimeException = 1666513645, then generate this over forcedTemplateParsing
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks for the patch, @Neowar - but this problem should not be fixed by suppressing the exception. The problem is that on some compositions of middlewares, VHS's asset inclusion somehow ends up _before_ `typo3/cms-frontend/prepare-tsfe-rendering` which is where the TypoScript gets set, which when missing causes the exception.

The correct solution is to use middleware configuration to ensure that VHS asset inclusion happens _after_ `typo3/cms-frontend/prepare-tsfe-rendering` (but before any other middlewares that measure/manipulate page content).

PR #1860, opened 2023-06-14
[BUGFIX] dirname(): Passing null to parameter #1

magicsunday:
PHP Runtime Deprecation Notice: dirname(): Passing null to parameter #1 ($path) of type string is deprecated in /var/www/public/typo3conf/ext/vhs/Classes/Service/AssetService.php line 761


NamelessCoder:
Solved meanwhile by other commit 2390e8066bda222d1a862192b9c3982ee76190bc

Issue #1905, opened 2024-04-23
BUG: Exception when "<v:page.rootline>" is used and pageUid doesn't exist


Oktopuce:
**I have checked that the bug exists in the `dev-development` branch**
Yes.

**I have checked that there are no already open issues or recently closed issues about this bug**
Yes.

**Describe the bug**
When you're trying to get the root line of a page that doesn't exist, a `PageNotFoundException` exception is thrown:

` (1/1) #1343589451 TYPO3\CMS\Core\Exception\Page\PageNotFoundException `

**To Reproduce**
1. Insert in a Fluid Template, a bloc like this:
```
<v:page.rootline pageUid="{linkNotInDb}" as="theRootLine">
    ...
</v:page.rootline>

```
with `{linkNotInDb}` having a page Uid that doesn't exist in DB.

2. Display the page with this code in frontend
3. See exception

**Expected behavior**
No `PageNotFoundException` exception should be thrown

**Additional context**
In file **vhs/Classes/Service/PageService.php**, method **getRootLine()** the call to  `$rootLineUtility->get()` must be surrounded by `try{} catch()`

```
try {
    /** @var RootlineUtility $rootLineUtility */
    $rootLineUtility = GeneralUtility::makeInstance(RootlineUtility::class, $pageUid);
    $rootline = $rootLineUtility->get();
} catch (PageNotFoundException $e) {
    return [];
}
```

NamelessCoder:
I would argue that in this case it's desirable to get an exception and undesirable to suppress all warning/error output when a non-existing page is passed. Otherwise you're just left clueless about why you get an empty root line.

I suggest using `v:try` around the root line extracting ViewHelper - this also lets you render an alternative when this exception is raised.

NamelessCoder:
I made the decision to reject this - it's undesirable to have this API suppress warnings about unreadable pages. See https://github.com/FluidTYPO3/vhs/pull/1906#issuecomment-2236564156.

PR #1906, opened 2024-04-23
Bugfix/1905

Oktopuce:


NamelessCoder:
I've made the decision to reject this, based on the argument I gave in https://github.com/FluidTYPO3/vhs/issues/1905#issuecomment-2104359749. It is undesirable to have this API silently suppress the warnings about unreadable page records - suggested alternative is to use `try` (in templates, `v:try`) and add a fallback case if you need to use ViewHelpers or API  where you are unsure if the page exists and need to tolerate cases where it doesn't. Just like you'd need to in any of your own code when using the core's RootLineUtility.

Issue #1902, opened 2024-03-13
BUG: no request object in v:render.template


queosailor:
**I have checked that the bug exists in the `dev-development` branch**
No

**I have checked that there are no already open issues or recently closed issues about this bug**
Yes

**Describe the bug**
If a template is included with v:render.template, no request object is available to the viewhelpers of the template.
e.g. f:translate is then no longer able to determine the site language
Also the f:email.link viewhelper is not able to spamprotect email adresses.

I Added 2 patches which solve these Problems:

[add_request2.patch](https://github.com/FluidTYPO3/vhs/files/14590590/add_request2.patch)
[add_request.patch](https://github.com/FluidTYPO3/vhs/files/14590591/add_request.patch)

PR #1907, opened 2024-05-25
[BUGFIX] Assign resolved Request in TemplateViewHelper

christophlehmann:
With this, for example the f:link.action() ViewHelper can be used in the rendered template

Resolves: #1902

### DIFF
--- a/Classes/ViewHelpers/Render/TemplateViewHelper.php
+++ b/Classes/ViewHelpers/Render/TemplateViewHelper.php
@@ -8,6 +8,7 @@ namespace FluidTYPO3\Vhs\ViewHelpers\Render;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use FluidTYPO3\Vhs\Utility\RequestResolver;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 
 /**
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks for the patch @christophlehmann - I should note that this will probably only affect TYPO3v12 since older versions do not have the `setRequest` method on View classes. But that's good enough for me since nobody else has reported this as a problem and I'm sure you already use v12 (since otherwise, your change would have triggered a fatal error).

haraldglaser:
This patch leads to an error if the template containing the viewhelper does not contain the necessary information for
RequestResolver::resolveRequestFromRenderingContext to resolve a request.

My usecase: this viewhelper is part of a custom backend preview (https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/ContentElements/CustomBackendPreview.html#ConfigureCE-Preview-PageTSconfig).
This now fails with message "Unable to resolve request from RenderingContext" thrown in RequestResolver::resolveRequestFromRenderingContext.
I think it's a legitimate use to include this viewhelper in a template which itself has no request attached. In this case in my opinion it shouldn't throw an error but proceed without setting a request on the targeted view.

NamelessCoder:
@haraldglaser Can you convert this comment to a new issue so it doesn't get lost? This is a pull request and can't be reopened, plus your use case is kind of different from the purpose of this patch and associated issue.

For the record: while you can technically operate this ViewHelper in a request-less context it certainly has limitations in terms of what _else_ you can do inside such a template. But I'm sure you're aware.

Issue #1901, opened 2024-02-20
TYPO3 13 compatibility


simonschaufi:
Please make this extension compatible with TYPO3 13

Pravin0711:
is there any news for v13?

NamelessCoder:
In general about support for new TYPO3 versions (and I'm sure this applies to basically every community extension) there are three main ways that new support gets added:

1. Sponsorships, usually either by way of the author (myself in this case) or a user working for a client that needs it and the client agrees to sponsor it.
2. Contributions, usually in the form of a user themselves needing the support and deciding to make the necessary changes and contribute them back as a pull request.
3. Spare time, where the author essentially pays with their own time/money to add the support.

In the specific case of VHS (and Flux) and TYPO3v13:

1. Hasn't happened yet. The overwhelming majority of users tend to stick with LTS versions and only when a new major LTS version is close (this isn't the case yet for v13) do they evaluate a possible need to sponsor the work to introduce the new support.
2. Hasn't happened yet, but would obviously be appreciated if a user does this.
3. I currently don't have the time or need - and right now I'm the only person maintaining the project.

jensscherbl:
Hi @NamelessCoder and @thomasrawiel 

- How far do you think is this extension away from compatibility with TYPO3 13 at the moment?
- What budget would be necessary for a Sponsorship to make this happen within the next two month (currently evaluating for a client project)?

NamelessCoder:
Hi @jensscherbl,

It's difficult to say (usually is with estimates for a task that's this broad). At the current time I'm aware of two major incompatibilities:

* Access to TSFE and properties on TSFE. Difficulty here is needing to maintain compatibility with 10.4 all the way through 13.4.
* Anything that builds links, which is quite a few things throughout the code base.

I am certain there are others but these are the ones that break my (admittedly limited) testing scenarios with hard exceptions.

A very rough estimate on those two (plus updates for the multiple test cases that those changes would definitely break) would probably be around 1600 EUR. An additional 800 EUR must be planned to add 13.4 builds and for test adjustments outside of those two major incompatibilities. I would also be a bit conservative and expect at least another 800-1200 EUR for issues that I'm currently not aware of - although those issues may not be complete deal-breakers and could possibly be skipped, allowing a partial compatibility release with some ViewHelpers disabled until they can be made compatible.

[...]

PR #1917, opened 2024-10-14
TYPO3 13 compatibility

thomasrawiel:
Here's my attempt for (beginning) TYPO3 13 compatibility

- added version check for addRootlineFields (obsolete in v13)
- Added a TSFE Utitlity class with version checks to try to substitute references to $GLOBALS['TSFE'] (Most TSFE members marked internal or read-only) while keeping compatiblity with old TYPO3 versions

related to #1901

### DIFF
--- a/Classes/Middleware/AssetInclusion.php
+++ b/Classes/Middleware/AssetInclusion.php
@@ -24,7 +24,7 @@ class AssetInclusion implements MiddlewareInterface
 
         /** @var AssetService $assetService */
         $assetService = GeneralUtility::makeInstance(AssetService::class);
-        $assetService->buildAllUncached([], $GLOBALS['TSFE'], $contents);
+        $assetService->buildAllUncached([], $request->getAttribute('frontend.controller', null) ?? $GLOBALS['TSFE'], $contents);
 
         if ($contentsBefore === $contents) {
             // Content is unchanged, return the original response since there is no need to modify it, or the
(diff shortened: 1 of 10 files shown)


nwm-dev:
**Unclosed '{' on line 164 does not match ')'**

in /vhs/Classes/ViewHelpers/Resource/AbstractImageViewHelper.php line 165

New Code:

$typoscript = (new TsfeUtility())->getTyposcriptSetupArray();

nwm-dev:
Fixed an Issue, where the `v:condition.string.contains` ViewHelper calls a removed function `evaluateCondition` in /vendor/fluidtypo3/vhs/Classes/ViewHelpers/Condition/String/ContainsViewHelper.php on line 32

affected line:
`protected static function evaluateCondition($arguments = null)`

fixed line:
`public static function verdict(array $arguments, RenderingContextInterface $renderingContext)`

TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface has to be added to "use"-definition.

NamelessCoder:
Over the past couple of weeks, fixes for the issues addressed in this PR have been solved through other means and with backwards compatibility to v10 preserved. Sorry to close this issue and with it discarding your work - but don't let that stop you from making PRs in the future. Keep in mind though the things I mention in my message above. Most importantly: make small changes and one thing changed per PR. Bulk PRs like this will never be merged ;)

PR #1933, opened 2025-02-05
[BUGFIX] Use addRootLineFields on v12.4

cweiske:
Commit "[BUGFIX] Avoid addRootLineFields on v13" prevented loading rootline fields on TYPO3 v12.4.26,
leading to a missing "nav_hide" field needed by the breadcrumb viewhelper.

We correctly check that it's a TYPO3 version less than 13, instead of less-or-equal than 12.4.

Related: 3b9720203c1df47fa943c01061f35116f93954ff


NamelessCoder:
Ah, d'oh. Of course that version constraint would exclude anything above `12.4.0`. Good catch, thanks!

PR #1934, opened 2025-02-14
BUGFIX: trim only if not empty

chrisi51:
PHP Runtime Deprecation Notice: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/vendor/fluidtypo3/vhs/Classes/ViewHelpers/Format/TrimViewHelper.php line 45


NamelessCoder:
Thanks for the patch @chrisi51. I've replaced the else condition with string casting and added it also to the other occurrence of trim. It's perfectly fine to trim null, boolean false or integer zero (which also reports as "empty") as long as it is cast to a string first. Plus, the warning would also potentially happen in the other case because `$content` might also be null in that case.

NamelessCoder (approved):

chrisi51:
yep, that's a better approach =)

PR #1940, opened 2025-03-17
[BUGFIX] Avoid relying on PageArguments in AssetService

smichaelsen:
The routing request argument isn't necessarily a PageArguments instance.

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -16,6 +16,7 @@ use Psr\Log\LoggerInterface;
 use TYPO3\CMS\Core\Cache\CacheManager;
 use TYPO3\CMS\Core\Log\LogManager;
 use TYPO3\CMS\Core\Routing\PageArguments;
+use TYPO3\CMS\Core\Routing\RouteResultInterface;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hey @smichaelsen - code of course looks just fine, but I haven't checked so maybe you know: if the `routing` attribute is *not* an instance of `PageArguments` (or a subclass of it), what else might it be and in which case would it be non-`PageArguments`?

I'm also wondering if this should result in a more explicit error like a specific exception, if the instance is an unexpected type, because not being able to see the page UID might prevent the AssetService from working correctly in many other ways (TS overrides, among others).

smichaelsen:
@NamelessCoder Thanks for taking a look.

In a regular TYPO3 request the routing attribute is set to a `SiteRouteResult` by the `SiteResolver` middleware and later to `PageArguments` by the `PageResolver`. So you'll always have both within a request.

The vhs AssetService is normally invoked in the context of the TYPO3 `RequestHandler` at which point you can expect a `PageArguments`, which you did and it normally works.

Where I'm experiencing difficulties is, that I'm operating in an API endpoint, where I do have a Site but not a page. I use TSFE and that triggers the vhs `AssetManager`. You can argue that I'm the one that's in muddy waters for using TSFE without `PageArguments`, but I still think it would be sane for vhs not to rely on `PageArguments` here.

NamelessCoder:
Thanks @smichaelsen - that makes perfect sense and an exception would not be a good thing for your use case.

One request though: I would like to preserve the method signature of `readPageUidFromContext` to keep the `:int` return type and have it return `0` in case the `routing` attribute isn't a `PageArguments`. Technically, a method signature change would constitute a breaking change (it being `protected` and all). I doubt anyone is overriding it but even so, we can avoid changing the signature with no negative effects. If you change that, I'll merge and tag a new release :)

smichaelsen:
@NamelessCoder Thank you, I altered my PR as requested. Thanks for keeping a close eye on avoiding breaking changes!

NamelessCoder (approved):

NamelessCoder:
Thanks! 7.1.1 is tagged now, if you want to update your own project(s).

PR #1945, opened 2025-04-21
Fix database deadlock exception in AssetService::getTypoScript

adamkoppede:
If the database caching backend is used for the "vhs_main" cache, database deadlocks occur if there are many concurrent requests to a page with uncached content. A full reproduction example is available on GitHub: https://github.com/adamkoppede/reproduce-deadlock-with-fluidtypo3-vhs-7-1

The deadlock occurs in the [`EXT:core Typo3DatabaseBackend::set() method`](https://github.com/TYPO3/typo3/blob/v13.4.9/typo3/sysext/core/Classes/Cache/Backend/Typo3DatabaseBackend.php#L83).  I created an issue in the upstream TYPO3 bug tracker: https://forge.typo3.org/issues/106593

Currently `Typo3DatabaseBackend::set()` is called four times per HTTP request. The first commit in this change set extends the `static::$settingsCache` so it contains `plugin.tx_vhs` fully, instead of just `plugin.tx_vhs.settings`; this reduces the `Typo3DatabaseBackend::set()` calls down to one per request. The second commit removes the `Typo3DatabaseBackend::set()` call for pages with uncached content, as they cannot be served from cache.

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -49,7 +49,7 @@ class AssetService implements SingletonInterface
     protected $cacheManager;
 
     protected static bool $typoScriptAssetsBuilt = false;
-    protected static ?array $settingsCache = null;
+    protected static ?array $typoScriptCache = null;
     protected static array $cachedDependencies = [];
     protected static bool $cacheCleared = false;
 
@@ -145,15 +145,15 @@ class AssetService implements SingletonInterface
      */
     public function getSettings(): array
     {
-        if (null === static::$settingsCache) {
-            static::$settingsCache = $this->getTypoScript()['settings'] ?? [];
-        }
-        $settings = (array) static::$settingsCache;
-        return $settings;
+        return $this->getTypoScript()['settings'] ?? [];
     }
 
     protected function getTypoScript(): array
     {
+        if (static::$typoScriptCache !== null) {
+            return static::$typoScriptCache;
+        }
+
         $cache = $this->cacheManager->getCache('vhs_main');
         $pageUid = $this->readPageUidFromContext();
         $cacheId = 'vhs_asset_ts_' . $pageUid;
@@ -192,6 +192,7 @@ class AssetService implements SingletonInterface
             $this->cacheManager->flushCachesByTag($cacheTag);
         }
 
+        static::$typoScriptCache = $typoScript;
         return $typoScript;
     }
 
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -50,6 +50,7 @@ class AssetService implements SingletonInterface
 
     protected static bool $typoScriptAssetsBuilt = false;
     protected static ?array $typoScriptCache = null;
+    protected static bool $currentlyBuildingCacheable = true;
     protected static array $cachedDependencies = [];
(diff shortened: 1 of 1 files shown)


adamkoppede:
Tests pass locally with `composer update --with typo3fluid/fluid:4.1.0` on my machine. This is the version used in the latest GitHub pipeline for the `development` branch: https://github.com/FluidTYPO3/vhs/actions/runs/14401105773/job/40386878009

The current `development` branch (`v7.1.1`) also fails with `composer update --with typo3fluid/fluid:4.1.1` locally. The update contains two changes related to condition view helpers: https://github.com/TYPO3/Fluid/releases/tag/4.1.1

Classes/Service/AssetService.php
```diff
@@ -50,6 +50,8 @@ class AssetService implements SingletonInterface
 
     protected static bool $typoScriptAssetsBuilt = false;
     protected static ?array $typoScriptCache = null;
+    protected static array $pagesWithSavedTypoScript = [];
```
NamelessCoder, line 53: AssetService should never be called on to generate assets for multiple pages in the same request. So we shouldn't need a multi-valued "already built" indicator. And we already have $typoScriptCache which indicates if TS has already been fetched (and possibly saved). Continues below.

Classes/Service/AssetService.php
```diff
@@ -151,26 +162,27 @@ public function getSettings(): array
     protected function getTypoScript(): array
     {
         if (static::$typoScriptCache !== null) {
+            $this->saveTypoScriptIfNecessary(static::$typoScriptCache);
```
NamelessCoder, line 165: This method should early-return if the TS has already been loaded, which means the TS will not be saved again after the first hit. And it doesn't seem right to save the TS in the early return (TS wouldn't exist in this property, unless it was already loaded and therefore already saved).

NamelessCoder (commented):
I think in general this very much looks like a job for a lock (semaphore or file ideally) rather than a range of workarounds. Using TYPO3's locking mechanism also allows users to configure how this locking happens (good for distributed systems).

And I guess it would be OK to make the thread wait for a lock release (maybe adding a maximum wait time).

There may also be a bug or unforeseen circumstance in how the AssetService uses the internal cache and writes to the cache, e.g. a missing condition to prevent cache writing when cache has already been written (which should not need new has-written tracking, see review comments).

But a lot of the code seems unnecessary or I don't understand the need for it. For example, keeping track of multiple page IDs, tracking whether the end result is cacheable, and storing those things in static variables.

Classes/Service/AssetService.php
adamkoppede: Yes, this isn't necessary if we assume that at most one page is processed per PHP invocation. I removed it.

Classes/Service/AssetService.php
adamkoppede: Yes, this isn't necessary if we assume that all `AssertService::buildAll()` calls within a PHP invocation either contain a cacheable or un-cacheable TSFE. If the same PHP invocation contains a mix, for example first call un-cached second call cached, this path would have been hit. I removed it.

NamelessCoder:
Asset feature removed in upcoming version.

PR #1952, opened 2025-06-23
[BUGFIX] rename Operation not permitted during cache flush

adamkoppede:
Remove the duplicate rename around the writes in
AssertService::clearCacheCommand(). AssetService::writeFile() itself is already implemented in an atomic fashion by using rename internally. Additionally, the use of GeneralUtility::tempnam() removed in this change triggered the following Exception in certain TYPO3 deployments when flushing the cache:

    PHP Warning:
    rename(
        /var/www/var/transient/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css.8B47TF,
        /var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css
    ): Operation not permitted
    in /var/www/public/typo3conf/ext/vhs/Classes/Service/AssetService.php line 757

Deployments are affected if they satisfy the following two conditions:

- TYPO3 cli and php-fpm are ran by different users, and
- "var/transient" and "public/typo3temp/assets" are located on different mount points.

If some existing cache file is owned by the other user, the utime(2) system call issued for the \touch() PHP function call fails with EPERM (Operation not permitted). The rename(2) system call issued for the \rename() PHP function call in the fallback path always fails with EXDEV (Invalid cross-device link). The PHP interpreter implicitly falls back to an alternative rename [1] that doesn't use rename(2) and is roughly equivalent to \fopen('w'), \fwrite(\file_get_contents($orig)), \chown() and \chmod(). But the \chown() and \chmod() operations cannot succeed because we aren't the owner of the file, as otherwise the utime(2) system call in the \touch() PHP function call would have succeeded. The same is true for the final utime(2) in the \touch() PHP function call which also just silently fails if PHP warnings are disabled. The following are the relevant system calls collected by inspecting php-fpm with strace:

    access("/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", F_OK) = 0
    utime("/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", {actime=0, modtime=0}) = -1 EPERM (Operation not permitted)
    rename("/var/www/var/transient/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css.nIoApp", "/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css") = -1 EXDEV (Invalid cross-device link)
    openat(AT_FDCWD, "/var/www/var/transient/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css.nIoApp", O_RDONLY) = 17
    openat(AT_FDCWD, "/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 18
    mmap(NULL, 20698, PROT_READ, MAP_SHARED, 17, 0) = 0x749df6db6000
    write(18, "/*!*****************************"..., 8192) = 8192
    write(18, "adding-top:5px}@media(min-width:"..., 8192) = 8192
    write(18, "in-top--xs{margin-top:15px}@medi"..., 4314) = 4314
    munmap(0x749df6db6000, 20698) = 0
    close(17)                  = 0
    close(18)                  = 0
    chown("/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", 83, 86800513) = -1 EPERM (Operation not permitted)
    chmod("/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", 0100600) = -1 EPERM (Operation not permitted)
    unlink("/var/www/var/transient/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css.nIoApp") = 0
    access("/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", F_OK) = 0
    utime("/var/www/public/typo3temp/assets/vhs/vhs-assets-09d7f183856cad2bc998f81e9e0660f4.css", {actime=0, modtime=0}) = -1 EPERM (Operation not permitted)

The usage of GeneralUtility::tempnam() over \tempnam() was introduced in bb5ae3f8 ("[BUGFIX] Write temp file within TYPO3 project path") to try to fix an issue where TYPO3 doesn't apply the correct permissions to the file due to being outside the project path. The earlier \tempnam() as well as the existing \tempnam() inside AssertService::write() explicitly pass in the directory of the target file. So if
GeneralUtility::fixPermissions() does nothing due to failing the GeneralUtility::isAllowedAbsPath() check it is correct in the sense that it would behave the same if it were called on the target file. Except that this wasn't true for symlinks inside the TYPO3 project pointing outside the TYPO3 project before this change, because \tempnam() expands symlinks. Configuring $GLOBALS['TYPO3_CONF_VARS']['BE']['lockRootPath'] on a per-file basis is no longer possible as of a70f34cbe14 ("[!!!][SECURITY] Enforce absolute path checks in FAL local driver"), which was ported to all TYPO3 branches.

[1]: https://github.com/php/php-src/blob/php-8.4.8/main/streams/plain_wrapper.c#L1327-L1376
Fixes: bb5ae3f8 ("[BUGFIX] Write temp file within TYPO3 project path")

### DIFF
--- a/Classes/Service/AssetService.php
+++ b/Classes/Service/AssetService.php
@@ -752,9 +752,7 @@ class AssetService implements SingletonInterface
         foreach ($assetCacheFiles as $assetCacheFile) {
             if (!@touch($assetCacheFile, 0)) {
                 $content = (string) file_get_contents($assetCacheFile);
-                $temporaryAssetCacheFile = (string) GeneralUtility::tempnam(basename($assetCacheFile) . '.');
-                $this->writeFile($temporaryAssetCacheFile, $content);
-                rename($temporaryAssetCacheFile, $assetCacheFile);
+                $this->writeFile($assetCacheFile, $content);
                 touch($assetCacheFile, 0);
             }
         }
@@ -778,6 +776,13 @@ class AssetService implements SingletonInterface
                 1733258066
             );
         }
+        if (realpath(dirname($file)) !== dirname($tmpFile)) {
+            throw new \RuntimeException(
+                "Failed to write file {$file}: No new files can be created in directory",
+                1750681587
+            );
+        }
+        $tmpFile = dirname($file) . '/' . basename($tmpFile);
         GeneralUtility::writeFile($tmpFile, $contents, true);
         if (@rename($tmpFile, $file) === false) {
             $error = error_get_last();


NamelessCoder:
I noticed we still have this in the writeFile method:

```
$tmpFile = @tempnam(dirname($file), basename($file));
```

Which I honestly thought we'd replaced with GU::tempnam, but I see that this was only done in the context of cache clearing.

Wouldn't the solution be to get rid of this `tempnam` call and use the GU wrapper for that?

It's fine to (also) throw an error if the `rename` function call fails, but I don't think we should be erroring out this way if the target folder is a symlink. As long as we're able to fix the permissions of the written file, which we would be if that is all done on the temporary file written by GU::tempnam, and are able to move the file to the target folder, we wouldn't need to care about symlink or not.

(Adding those paths to lockRootPath if they are outside project root might still be necessary, but that works just fine)

adamkoppede:
> I noticed we still have this in the writeFile method:
> 
> ```php
> $tmpFile = @tempnam(dirname($file), basename($file));
> ```
> 
> Which I honestly thought we'd replaced with GU::tempnam, but I see that this was only done in the context of cache clearing.
> 
> Wouldn't the solution be to get rid of this tempnam call and use the GU wrapper for that?

What is the problem with `@tempnam()` that `GU::tempnam` fixes? Why are we replacing `@tempnam()` with `GU::tempnam`?

The second paragraph in the change description explains why we cannot use `GU::tempnam` in this one project, but I'll try another way of wording of it:

The TYPO3 deployment uses two separate hosts with each of them having their separate php-fpm + apache and host-local file system. Only "public/typo3temp", "public/uploads" and "public/fileadmin" are shared between them using nfs. This means that "var/transient", which is the directory `GU::tempnam` creates its files in, and "public/typo3temp/assets", from which apache serves the EXT:vhs assets from, are on different mount points. This means that a `rename(2)` system call always fails with EXDEV (Invalid cross-device link).

The TYPO3 deployment also uses different users for executing php-fpm and TYPO3 cli. This means that existing files in "public/typo3temp/assets" are either owned by the php-fpm user (if they were created during a user-triggered page load) or by the TYPO3 cli user (if they were created during cache pre-warming).

Now this happens when we encounter a file that is owned by a different user in `AssetService::clearCacheCommand()` loop: The `@touch()` php function call fails because we aren't the owner of the file. The `@rename()` php function call inside `AssetService::writeFile()` first fails to use the `rename(2)` system call with EXDEV (Invalid cross-device link). The php runtime than transparently falls back to directly writing to the target destination. After the php runtime copied the file contents, it fails to copy the owner and permission information of the file with error EPERM (Operation not permitted). This error is never reported to the php application; the destination file just silently remains with the wrong permissions and the `@rename()` php function call completes successfully. The second `@touch()` php function call inside the loop fails with EPERM (Operation not permitted) because we're still not the owner of the file.

> It's fine to (also) throw an error if the rename function call fails, but I don't think we should be erroring out this way if the target folder is a symlink. As long as we're able to fix the permissions of the written file, which we would be if that is all done on the temporary file written by GU::tempnam, and are able to move the file to the target folder, we wouldn't need to care about symlink or not.

I'm not aware of any place where we error out if the target folder is a symlink.

I guess you mean the following check I introduced:

```php
if (realpath(dirname($file)) !== dirname($tmpFile)) {
    throw new \RuntimeException(
        "Failed to write file {$file}: No new files can be created in directory",
        1750681587
    );
}
```

The reason for this is that `@tempnam()` falls back to the system's temporary directory if no temporary file could be created in the directory we specified. But if we cannot write into this directory we also shouldn't be able to rename the temporary file to its target destination. We need to check against the resolved destination directory name because `@tempnam()` implicitly resolves symlinks.

> (Adding those paths to lockRootPath if they are outside project root might still be necessary, but that works just fine)

When testing this patch I didn't notice any issues when I left lockRootPath empty in a testing environment where "public/typo3temp" is symlinked outside the typo3 project directory. This is because the PHP code explicitly avoids implicitly resolving symlinks in the `@tempnam()` php function call using the following snippet:

```php
$tmpFile = dirname($file) . '/' . basename($tmpFile);
```

Since "public/typo3temp/assets" is inside the public directory `GU::isAllowedAbsPath()` always succeeds.

NamelessCoder:
> What is the problem with `@tempnam()` that `GU::tempnam` fixes?

The location where the file is written. System tempnam will create the file in the temporary directory of PHP (almost certainly outside TYPO3 project dir) which means `GU::writeFile` can't adjust permissions on the file. Yes, we can pass the desired directory to `tempnam`, but that doesn't guarantee against the exact same problem: causing the file to be written outside TYPO3 project folders.

We're wanting to use `GU::tempnam` because this allows the user to configure where these temp files are written, through setting the `var` path. We _want to use the temporary path_ instead of writing directly to the final path. While it's true that VHS does write to `typo3temp` currently, which "smells" like a temporary files path, this is only done for legacy reasons (10.4 / 11.5 support) and is almost certain to be changed to something inside `_assets` in a future version that drops legacy support.

So if we don't use the core's temp mechanisms we're just bound to run into these problems again in the future, since that would treat a location under `_assets` as a temp dir - and that is definitely not desired.

> I'm not aware of any place where we error out if the target folder is a symlink.
> I guess you mean the following check I introduced:

Yes, that's the code in question. It's checking if `realpath` differs from implied path - so it's not checking if the file can be renamed (which would mean checking return from `@rename`). At the very least, it's an akward backwards way of checking writability.

--------------------------------------------

Long story short: all of this seems to boil down to what actually happens in the cache flushing method. Rather than reporting inability to `touch` a file as an error (which it imho certainly is: users that need to clear cache must be able to write - in this case, `touch()` - those generated asset files), https://github.com/FluidTYPO3/vhs/commit/93526155911499d8feb6f3032d4b233da29f4845 added a workaround that recreates the file (with new ownership). Which may have been added because further up the chain, `GU::writeFile` failed to adjust permissions because of the location of the file while it was created with `tempnam` and not `GU::tempnam`.

At this point it may be more correct to:

* Assume and require that any file written by any PHP process is owned by a user and has permission flags which also allows access for any other user which needs to access the file (CLI deployer, scheduler, whatever). All of these aspects are configurable: ownership group, creation permissions/umask.
* Leave it up to the system admin to set the right ownership/permission configuration through `TYPO3_CONF_VARS`.
* If `touch` fails, throw an error reporting this as an actual error stemming from user/permissions config (no more file recreation workaround which just changes the owner and moves the permission problem to the opposite side).
* Use `GU::tempnam` also in `AssetService::writeFile` to ensure file is written to a place where permissions can be set according to TYPO3 configuration.
* Which implies that `public/typo3temp` should be located on the same filesystem as the `var` folder (the `typotemp` folder can be a symlink to there, thus preventing EXDEV).
* As last stop-gap measure, introduce a PSR-14 even that can be dispatched from `AssetService::writeFile` which allows third parties to take over all asset file creation responsibilities in case their design decisions make the above impossible.

Doing this means we no longer hide the actual reasons why files can't be modified by one user if they're created by another user - which is a problem that _should be fixed by configuring TYPO3 to use appropriate owner and permissions for files it creates_ rather than worked around.

I would rather not have VHS take even more responsibility for working around various edge cases, but I'd be happy to remove existing workarounds, require proper configuration, throw precise errors, and provide options to externally replace the file writing logic.

adamkoppede:
I'm closing this issue. Not wanting to support `public/typo3temp` and `var` being on different mount points is absolutely reasonable to me as this is quite a niche setup. Thanks for your time and for maintaining this extension.

---

> We're wanting to use GU::tempnam because this allows the user to configure where these temp files are written, through setting the var path.

This is what this patch conflicts with. I see that this may be desirable, as for example you'd have a single place to look for forgotten temporary files to clean them up, but it is impossible to provide across different mount points.

---

> The location where the file is written. System tempnam will create the file in the temporary directory of PHP (almost certainly outside TYPO3 project dir) which means GU::writeFile can't adjust permissions on the file. Yes, we can pass the desired directory to tempnam, but that doesn't guarantee against the exact same problem: causing the file to be written outside TYPO3 project folders.

> Yes, that's the code in question. It's checking if realpath differs from implied path - so it's not checking if the file can be renamed (which would mean checking return from @rename). At the very least, it's an akward backwards way of checking writability.

The PHP `tempnam()` function always tries to create the file in the specified directory first. Only if that fails, it'll go ahead and create the file in the system temporary directory. The `realpath(dirname($file)) !== dirname($tmpFile)` check is to detect if the earlier `tempnam()` PHP function call created the file in the specified directory or if it fell back to the system temporary directory. If it is the later, the \RuntimeException is thrown early as we most likely won't be able to `rename()` any file into there if creating a file there failed. With that we never `GU::writeFile()` on files inside the system temporary directory.

---

`touch()` only works if you're the owner of the file. It is unaffected by configuring `$GLOBALS['TYPO3_CONF_VARS']['SYS']['fileCreateMask']` and `$GLOBALS['TYPO3_CONF_VARS']['SYS']['folderCreateMask']`. So the fallback path where we create a new file if `touch()` fails is necessary to support having multiple users who share their user group. This is supported by TYPO3 core by setting the two options mentioned to `0664` and `2775` respectively. I don't see how one could introduce a TYPO3 configuration to avoid the fallback path.

NamelessCoder:
Thanks - I had managed to forget that `touch` won't allow modification time change when you're in the same _group_ as the owner. You actually have to _be_ the owner.

Given that this is becoming more and more normal to have two different users involved and the non-http user needing access to flushing caches, it's probably best to come up with a different method of expiring the files. I'll think about this.

I'll also try to get some PSR-14 events in there to let you replace the file writing altogether and wrangle it to a form where you're keeping within the same file system boundary - without sacrificing it being an atomic operation.

PR #1955, opened 2025-08-12
Allows cache configuration to be overridden

nalmar:
Allows vhs cache configuration to be overridden by additionnal.php while keeping extension defaults for non-overridden values.

if my case, I add 'backend' = 'TYPO3\CMS\Core\Cache\Backend\RedisBackend' and a custom defaultLifetime.

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -13,26 +13,24 @@
         $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['t3lib/class.t3lib_tcemain.php']['clearCachePostProc'][] = \FluidTYPO3\Vhs\Service\AssetService::class . '->clearCacheCommand';
     }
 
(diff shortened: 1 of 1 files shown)


NamelessCoder (changes requested):

NamelessCoder:
This has to be approached in a different way:

Extensions' cache configurations are always added after base configuration loads. As such, the cache configurations don't actually exist until the extensions' ext_localconf.php is loaded. The configurations added by extensions _should_ override any information previously set - that's simply how extensions _should_ work.

The only reason you can override the _core's_ cache configurations in `additional.php` is that the core ships the initial configuration in the huge `DefaultConfiguration.php` boilerplate. Extensions do not have this ability - this boilerplate is created and loaded far earlier than any extension loads.

To override the cache configurations (or indeed, every single other type of configuration contained in an extension) the right approach is to create an extension which has the extension being overridden as a dependency, then perform your cache configuration overrides in that extension's `ext_localconf.php` file. That ensures the processing order, and works for every extension.

PR #1956, opened 2025-09-01
[TASK] Implement backwards compatible base VH class

NamelessCoder:
Prevents deprecation log flooding without requiring massive API changes.


liayn:
Seems like this changed killed `v:render.uncache`

> Thu, 04 Sep 2025 09:36:35 +0200 [ALERT] request="d124be27c990a" component="TYPO3.CMS.Frontend.ContentObject.Exception.ProductionExceptionHandler": Oops, an error occurred! Code: 20250904093635450d5026- InvalidTemplateResourceException: Tried resolving a template file for controller action "Default->Default" in format ".html", but none of the paths contained the expected template file (Default/Default.html). No paths configured., in file /var/www/html/web/vendor/typo3fluid/fluid/src/View/TemplatePaths.php:617 - {"exception":"TYPO3Fluid\\Fluid\\View\\Exception\\InvalidTemplateResourceException: Tried resolving a template file for controller action \"Default->Default\" in format \".html\", but none of the paths contained the expected template file (Default/Default.html). No paths configured.

At least this happened, when we updated from 7.1.3 to 7.1.4

NamelessCoder:
Pretty sure this change isn't the one causing a problem with `v:render.uncache`. Can you try reverting commit 68d84ad4be3a6d578fbcda47ba336919af8a0567 to see if that solves the problem?

liayn:
Right, I clicked the wrong PR, sry. Yes,
https://github.com/FluidTYPO3/vhs/commit/68d84ad4be3a6d578fbcda47ba336919af8a0567#diff-38f5ddad84242dbb78f7a031ac97a0ba3df9c4e58476fa3007d5e88bbaf99948L117
is the culprit in our case.

PR #1964, opened 2026-01-16
[BUGFIX] Load server request from rendering context

### LINKED ISSUE #1937: BUG: content.resources.fal returns an empty array in version 7.1.0
**I have checked that the bug exists in the `dev-development` branch**
Yes

**I have checked that there are no already open issues or recently closed issues about this bug**
Yes

**Describe the bug**
`{v:content.resources.fal(field: 'someImageField')}` returns an empty array after upgrading to vhs 7.1.0. Up until version 7.0.7 it returned an array with all the image properties such as id, name, url etc, which could be used to display the image.

**To reproduce**
1. Create a flux template with the following field: `<flux:field.inline.fal name="someImageField" label="Some Image" />`
2. Put `<f:debug>{v:content.resources.fal(field: 'someImageField', record: record)}</f:debug>` in the frontend template part.
3. Create a new content element using the flux template and view it in the frontend.

**Expected behavior**
`f:debug` should output an array of arrays with the image name, id etc. as it did before.

**Additional context**
In the preview part of the template the ViewHelper still works as before. The difference is, that I'm adding a record parameter here, as described by the documentation: `{v:content.resources.fal(field: 'someImageField', record: record)}`
In fact, if I do this in the frontend template part too, it works again. According to the documentation it should work without the record parameter.

Is this a bug introduced by 7.1.0? Or are we exprected to now always provide the record parameter when using `v:content.resources.fal`?

cweiske:
Since commit 99129162418ae0170e66dfafc4cf55980c60a002
> [BUGFIX] Always fetch ContentObject from request in TYPO3v12+

the current content element is not available anymore in view helpers unless passed manually ("v:content.resource.fal(record: record)"). This leads to e.g. "v:content.resources.fal" not loading images anymore.

The current content object is part of the Request object. A new request object is created when a new content element is rendered. This is not $GLOBALS['TYPO3_REQUEST'] - the per-element request object gets passed to each view helper via the rendering context.

The rendering context API changed with TYPO3 v13[1], making it necessary to first try the attribute (TYPO3 v13+) and fall back to getRequest() in TYPO3v12.

[1] https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.3/Deprecation-104684-FluidRenderingContext-getRequest.html

Resolves: https://github.com/FluidTYPO3/vhs/issues/1937

----

I've successfully tested it with image content elements that use `v:content.resource.fal` on TYPO3 v12.

### DIFF
--- a/Classes/Traits/SourceSetViewHelperTrait.php
+++ b/Classes/Traits/SourceSetViewHelperTrait.php
@@ -96,7 +96,7 @@ trait SourceSetViewHelperTrait
         ?string $params = null,
         ?string $crop = null
     ): array {
-        $contentObject = ContentObjectFetcher::resolve($this->configurationManager);
+        $contentObject = ContentObjectFetcher::resolve($this->configurationManager, $this->renderingContext);
         if ($contentObject === null) {
             throw new Exception(static::class . ' requires a ContentObjectRenderer, none found', 1737808465);
         }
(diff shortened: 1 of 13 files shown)


cweiske:
Getting the Tests run under TYPO3 v13 is a hard task...

cweiske:
Will not continue this.

PR #1966, opened 2026-02-03
[BUGFIX] Fix LinkViewhelper outside of extbase context

### LINKED ISSUE #1943: BUG: Viewhelper LinkViewhelper in VHS 7.1.0 with TYPO3 12.4.28
**I have checked that the bug exists in the `dev-development` branch**
_Yes or no._
Yes

**I have checked that there are no already open issues or recently closed issues about this bug**
_Yes or no._
Yes

**Describe the bug**
When we use the viewhelper <v:page.link pageUid="{uid}"/> it return a error occurred in frontend.
Error return : 
`Exception: TYPO3\CMS\Extbase\Mvc\Web\Routing\UriBuilder::setRequest(): Argument #1 ($request) must be of type TYPO3\CMS\Extbase\Mvc\RequestInterface, TYPO3\CMS\Core\Http\ServerRequest given, called in /var/www/composer/sit_services/packages/vhs/Classes/ViewHelpers/Page/LinkViewHelper.php on line 236, in file /var/www/composer/sit_services/vendor/typo3/cms-frontend/Classes/ContentObject/ContentObjectRenderer.php:722`

**To Reproduce**
1. Create a link viewhelper in your html : <v:page.link pageUid="{uid}"/>
2. See error 

**Expected behavior**
Create a link in the frontend like it happens in older version of VHS example version 7.0.7

**Screenshots**

Image

Image

**Additional context**
Typo3 use : 12.4.28 and 12.4.27 in composer mode

RKlingler:
This resolves Resolves #1943.

Using <v:page.link pageUid="{uid}"/> outside of an extbase context - for example in a page template - triggers an error because the request object cannot be retrieved correctly. The responsible method in RequestResolver can return either RequestInterface (within extbase context) or ServerRequestInterface (outside of extbase context). The UriBuilder in the view helper class requires the type to be RequestInterface though.

I added a new method to RequestResolver, which always returns RequestInterface.


RKlingler:
I tried several times in recent months to find a solution to this that also works in TYPO3 10 and 11. The required types and the methods which provide them vary just too much from version to version.
I'll leave it at that. But I could have another go at it when support for older TYPO3 versions is dropped and in case the issue still persists at that point.

NamelessCoder:
No longer an issue in `development`, will be included in next release.

PR #1971, opened 2026-03-26
um die EXT: verwenden zu können

colaholiker:


NamelessCoder:
Hi @colaholiker - this seems reasonable enough, but I would like:

* The GeneralUtility class to be imported so you can make the method call with `GeneralUtility::getFileAbsFilename`.
* A check before calling this method that the input must start with `EXT:` before you call the class, so we don't break any current implementations.

PR #1974, opened 2026-05-27
Update for TYPO3 14.3

BenjaminBeck:
Hi NamelessCoder,

most of this was generated with AI, mainly ChatGPT 5.5.
  
Regarding support for older Versions:
I suggest support for Typo3 13.4 and 14.3 - each with PHP 8.2,8.3,8.4 and 8.5
( https://typo3.com/typo3-cms/development-roadmap/roadmap )

Starting this was not really planned... my time doing this task is pretty limited.
And i am not sure how good the AI work is. But i think the nature of this project with its small logical units and the big test suite could produce good results.

Help is appreciated:

👉 Please install this PR in a Dev-Env and report what breaks


Classes/Utility/ContextUtility.php
```diff
@@ -8,17 +8,26 @@
...
 
+use Psr\Http\Message\ServerRequestInterface;
 use TYPO3\CMS\Core\Http\ApplicationType;
 
 class ContextUtility
 {
     public static function isFrontend(): bool
     {
-        return ApplicationType::fromRequest($GLOBALS['TYPO3_REQUEST'])->isFrontend();
+        $request = $GLOBALS['TYPO3_REQUEST'] ?? null;
```
BenjaminBeck: shouldnt be $GLOBALS['TYPO3_REQUEST'] the fallback after using the request from ?

NamelessCoder:
I appreciate the effort, but I'm sorry: there is just no way that I can adequately review a pull request that modifies more than 7,000 lines of code in 451 files and includes AI-opinionated unnecessary changes in everything from metadata over build instructions to constructors, annotation and method return types - and I'm not passing a change that I cannot review and sign-off on.

For this to be considered it *has* to be cut down to size with every change that isn't absolutely necessary removed. The pretty and nice to have stuff can then be considered after the important stuff.

And I would strongly discourage letting an AI attempt this, because they *will* start going off in all kinds of directions as soon as they hit some blocker. They are not exactly known for coming up with elegant and small compatible changes...

